PR #158, opened 2019-07-16
Make wheel damage (almost) indepenent of vehicle velocity

STEngineer:


DonBruce64:
Your PR does what it says in the description, but I'm still confused as to *why* you are changing that code.  If you get hit by a truck going 50MPH you'll get a lot more pain than one going 10MPH...

STEngineer:
> Your PR does what it says in the description, but I'm still confused as to _why_ you are changing that code. If you get hit by a truck going 50MPH you'll get a lot more pain than one going 10MPH...

I'm sure `get hit` isn't equal to `get run over`

DonBruce64:
For the purposes of this code it is.  There is no code in the mod that detects if a player is hit by a vehicle except that which looks at the wheels.  The reason is that many vehicles, such as planes, have collision boxes that should not harm you if you touch them while the vehicle is moving.  Yes, getting smacked by a wing would hurt, but brushing along the side of the fuselage of a jet won't do a thing to you.  By removing the velocity from the wheel hit equation you effectively make the damage the same for getting hit/run over no matter if the vehicle is going 5MPH or 50MPH.

DonBruce64:
Closing due to the code change resulting in vehicles giving the same damage values no matter what speed they are going, and no further discussion from requester.

PR #172, opened 2019-09-20
Multiple changes

SteveTheEngineer:
The changes are:
- Add a **config option** to ignore vehicle speed on calculating damage caused by wheels
~~- Make any part weight 50kg by default (earlier it was only applicable to parts that are not crates or barrels)~~ (removed in commit [2b538d6](https://github.com/DonBruce64/MinecraftTransportSimulator/commit/2b538d6557bd6f89216baf37929ead0ca49223ec))
- Change HeavyItems to a proper (by my mind) system, so additional mass is added instead of doubling


DonBruce64:
Just want you to know that this PR is still on my to-do list.  Just trying to get the bugs and kinks out for the V15 release.

STEngineer:
Okay thank you

STEngineer:
> Additionally, the priority system seems kinda silly, as I don't see any reason people would ever have an entry in there twice or need to "prioritize" things as iron will only weigh what they say it to weigh.
Why not anyway?

SteveTheEngineer:
[b2b774](https://github.com/DonBruce64/MinecraftTransportSimulator/commit/b2bb774aae7b13b8b62875b683c15139153635e8)

PR #252, opened 2020-01-06
Russian translation update

StarterCraft:
So the Russian translation of ImmersiveVehicles coremod became very outdated now, and I've updated it for harmony. My translations're the better comparing to other, so I hope this will be added into the official mod file 😀


DonBruce64:
Looks good to me!  Will be in the next release!

PR #259, opened 2020-01-12
Add Simplified Chinese translation

HydrogenC:
Simplified Chinese (zh_CN) translation, manual and achievements excluded.


DonBruce64:
Looks good to me!

PR #277, opened 2020-01-21
More zh_CN translations

HydrogenC:
Also unify the semicolons.


DonBruce64:
Looks good to me!  I'll be moving the handbook translations into their own special item JSONs this update, so those will no longer be in the lang file, but on the bright side you'll get your name on the cover of a book!

HydrogenC:
I will try to translate the handbook after this update.

DonBruce64:
Rodger.  Handbooks were moved into their own .json file to use the new booklet code, so all you have to do to add yours is copy src/main/resources/assets/mts/jsondefs/booklets/handbook_en.json and re-name it to whatever suffix you like, and translate the text in "" in there.  See src/main/resources/assets/mts/jsondefs/booklets/handbook_ru.json for an example of one such translation.

HydrogenC:
I will translate it as soon as I have plenty of time.

PR #297, opened 2020-02-06
Updating manual handbook file

StarterCraft:
Replaced the ugly googletranslaty title of the handbook to a nice one.
Replaced MTS mentioning with IV mentions.
Updated the manual text due to the IV mechanics' changes.


DonBruce64:
Looks good to me!  Will have this in the next update.

PR #349, opened 2020-03-13
Turret rotation limit

chj990707:
Tripod Gun and Turret Gun now looks at JSON def for their rotation limits. if there is no value in JSON(that equals 0), they use default values. If the limit needs to be 0, JSON must have some value close enough to it.
Turrets can rotate around without limit if the minYaw is positive & maxYaw is negative.

Since the suggestion I made about separating the gun into shooting part on traversing part would take some time, this would do until I figure out how.


DonBruce64:
Back to you.  Pitch is inverted and the turret get spazzy when crossing the 180/-180 line.

DonBruce64:
Got the spazzy guns.  Just need pitch.

DonBruce64:
Awesome!  In this goes into the upcoming release!

PR #362, opened 2020-03-24
animal passengers

chj990707:
Leaded animals can be placed on the seat. They can be removed by Rclick, or lead Rclick(which instantly lead them). Animals are not rotated according to seat rotation.

### DIFF
--- a/src/main/java/minecrafttransportsimulator/vehicles/parts/PartSeat.java
+++ b/src/main/java/minecrafttransportsimulator/vehicles/parts/PartSeat.java
@@ -6,9 +6,15 @@ import minecrafttransportsimulator.jsondefs.JSONVehicle.VehiclePart;
 import minecrafttransportsimulator.packets.general.PacketChat;
 import minecrafttransportsimulator.vehicles.main.EntityVehicleE_Powered;
 import net.minecraft.entity.Entity;
+import net.minecraft.entity.EntityLiving;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.entity.player.EntityPlayerMP;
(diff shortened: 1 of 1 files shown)


DonBruce64:
So a few comments on this one.
1) when testing equality on objects (things that aren't integers, floats, etc.) you should use equals.  So, as an example:
heldStack.getItem() == Items.LEAD

Becomes:
Items.LEAD.equals(heldStack.getItem())
I'm not 100% sure why, but sometimes the Java system will return false for the == method but true for the .equals() method.  I blame funky Java design philosophy.


2) player.getHeldItemMainhand() will never be null.  At least not in 1.12.2+ versions.  That returns ItemStack.EMPTY if the player doesn't have anything, which can still have getItem() called on it.  That also won't return null.  Instead, it returns Items.AIR.  This is why you sometimes see Air as a crafting ingredient in MTS benches if stuff is wrong.

The whole idea behind this is to prevent NullPointerException errors from comparisons just like you are doing.  On that note, to compare to the lead you can either do Items.LEAD.equals(heldStack.getItem()), or invert it to heldStack.getItem().equals(Items.LEAD).  As a personal preference to keep me out of pain, I always structure my code where the "known present object" is the one that has the equals called on it.  If heldStack.getItem() is null, then heldStack.getItem().equals(Items.LEAD) will error out as you called a method on a null, while Items.LEAD.equals(heldStack.getItem()) will simply return false as you've asked if Items.LEAD is null (which it ain't).

Don't have to fix or change this code.  In fact, I know I still have a lot of itemStack != null code in my codebase.  But thought it'd be something I'd mention.

PR #383, opened 2020-04-12
Update en_US.lang

DrprofLuigi:
Added autopilot lang definition.


SteveTheEngineer:
Why 5 pull requests instead of 1?

PR #384, opened 2020-04-13
Add supercharger code to engines.

fsendventd:
Took me long enough...


DonBruce64:
Well, Git won't tell me the differences of everything, but I trusts ya.  Off it goes into the next release!

PR #387, opened 2020-04-14
Custom engine pitch, Custom gear shift, and tire squeal

BlazeWolf1122:
This allows custom gearshift/pitch by allowing MTS to first detect if there's a section for custom gearshift/pitch (provided by the JSON), then apply those pitches/gearshifts
If the section is false or doesn't exist, MTS will ignore it and do it's default stuff as if it never existed
Tire squeal is just adding a tire squealing sound to the section that causes burnouts and stuff
I dont exactly know how to use github for several PRs, so it's all in this one big peice


src/main/java/minecrafttransportsimulator/vehicles/parts/APartEngine.java
```diff
@@ -439,22 +439,35 @@ public void updateProviderSound(SoundInstance sound){
...
 				sound.pitch = (float) (0.35*(1 + Math.max(0, (RPM - engineStartRPM))/(definition.engine.maxRPM < 15000 ? 500 : 5000)));
-			}
+				}
 		}else if(sound.soundName.endsWith("_supercharger")){
 			if(!state.running && internalFuel == 0){
 				sound.stop();
 			}else{
-				//Pitch should be 0.35 at idle with a 0.35 increase for every 2500 RPM, as with the regular engine sound, but in addition the volume increases with RPM so that it's 1.0 at the engine's max safe RPM.
-				sound.volume = (float) RPM/definition.engine.maxRPM;
+				//Pitch should be 0.35 at idle, with a 0.35 increase for every 2500 RPM, or every 25000 RPM for jet (high-revving) engines by default.
```
fsendventd, line 463: You copied this directly from the engine section. Should match the comment I put in for how the supercharger sound is affected.

src/main/java/minecrafttransportsimulator/vehicles/parts/APartEngineGeared.java
```diff
@@ -20,13 +20,22 @@ public APartEngineGeared(EntityVehicleE_Powered vehicle, VehiclePart packVehicle
 	public void updatePart(){
 		super.updatePart();
 		//Do automatic transmission functions if needed.
+		//"!definition.engine.customShifter" If we don't do that first the MTS will break all engines that don't the custom shifter system
```
fsendventd, line 23: Word seems to be missing. Should be "don't use the custom shifter system".

src/main/java/minecrafttransportsimulator/vehicles/parts/APartEngine.java
fsendventd: You also forgot to implement the volume changing for superchargers back into your code section when you added custom pitches. That part is almost more important than the pitch, as the pitch of a supercharger is fairly constant but the volume changes significantly as the engine's speed changed.

src/main/java/minecrafttransportsimulator/vehicles/parts/APartEngineGeared.java
```diff
@@ -10,7 +10,7 @@
 
 public abstract class APartEngineGeared extends APartEngine{
 	public byte currentGear;
-
+	public float getLastGear;
```
DonBruce64, line 13: This isn't used anywhere in the file.  This should not exist.

src/main/java/minecrafttransportsimulator/vehicles/parts/PartGroundDeviceWheel.java
```diff
@@ -158,6 +158,7 @@ public void spawnParticles(){
 		if(skipAngularCalcs && this.isOnGround()){
 			for(byte i=0; i<4; ++i){
 				Minecraft.getMinecraft().effectRenderer.addEffect(new VehicleEffectsSystem.ColoredSmokeFX(vehicle.world, partPos.x, partPos.y, partPos.z, Math.random()*0.10 - 0.05, 0.15, Math.random()*0.10 - 0.05, 1.0F, 1.0F, 1.0F, 1.0F, 1.0F));
+				WrapperAudio.playQuickSound(new SoundInstance(this, MTS.MODID + ":" + "wheel_squeal"));
```
DonBruce64, line 161: This isn't a looping sound.  Looping sounds have an extra "true" parameter.  Also, you're going to spawn one of these every time the particle spawns, leading to intense lag.  What you should do is check whenever particles start spawning and set a boolean variable to the tune of tiresCurrentSqualing, or tiresSpinning, etc.  The moment you set that and it goes from false to true, you start the looping sound.  Then, you check it in the updateSounds() method.  Whenever that goes false, you stop the squealing sound.

src/main/java/minecrafttransportsimulator/vehicles/parts/APartEngineGeared.java
```diff
@@ -20,13 +20,22 @@ public APartEngineGeared(EntityVehicleE_Powered vehicle, VehiclePart packVehicle
 	public void updatePart(){
 		super.updatePart();
 		//Do automatic transmission functions if needed.
+		//"!definition.engine.customShifter" If we don't do that first the MTS will break all engines that don't use the custom shifter system
```
DonBruce64, line 23: Negative.  This parameter is un-needed if you replace it with:
definition.engine.upShiftRPM != null && definition.engine.downShiftRPM != null

src/main/java/minecrafttransportsimulator/vehicles/parts/APartEngine.java
```diff
@@ -439,22 +439,33 @@ public void updateProviderSound(SoundInstance sound){
 			if(!state.esOn && !state.hsOn){
 				sound.stop();
 			}else{
-				sound.pitch = (float) ((engineStartRPM*0.25F + RPM*0.75F)/engineStartRPM);
+				if (definition.engine.isCrankingNotPitched){
```
DonBruce64, line 442: Still VERY un-sure about static cranking pitch. Not that it can't be done, but that perhaps it shouldn't be done. Ian Malcolm would like to have a word with you, and I'd like you to ask Limit and DPL about this change.

PR #396, opened 2020-04-17
"Advanced" Custom gearing and engine sounds

BlazeWolf1122:
Use up to 3 custom sounds
Pitch modifiers
Volume modifiers
Sound Identification/name (Similar to how you identify sounds in the motorized horn section)
Modify Normal _running and _supercharger sound
Pitch modifiers
Volume modifiers
In total you have up to 5 "advanced" sounds to use
_Running
_Supercharger
Custom sound 1
Custom sound 2
Custom sound 3

Pitch/VolumeMin: the value that the pitch/volume of the engine starts at. This value can be negative (and clamps at 0 as you already know) to simulate things like a engine start/stop system in a hybrid where the engine kicks in at higher speeds but at lower speeds you hear the electric motor hum

Pitch/VolumeRev: the value that the pitch/volume of the engine reaches at max RPM. This value can be negative (and clamps at 0 as you already know) to simulate things like a engine start/stop system in a hybrid
The pitch Rev cannot be negative else the system will ignore it and pass it off and use the normal MTS sound and volume settings.The volume Rev can be negative though, which is useful for a engine idle

Cranking "un-pitched" for engine cranking sounds designed to maintain pitch (until the battery gets low)

UseCustomShifter: true false self explanatory
I had to keep this value because the suggestion you made either crashed the game if you used engines without custom gear shifting, or stopped those engines without custom shifting from shifting at all.
upShiftRPM/downShiftRPM: This tells MTS when to shift and shifts at the engine speed you tell it to. useful for giving smaller engines with gearing a large kick in performance, reducing fuel consumption, and increasing overall engine reliability, as well as adding a whole host of new ways to play with gearing in MTS engines to do all sorts of wacky stuff
Shift thresholds are made by replacing the gear ratios with the engine RPM you want to shift at, and removing the neutral and reverse. You must keep last gear though

If you have anything to tell me, please tell me in the MTS discord since i can't just casually pull up github to make changes and communicate


src/main/java/minecrafttransportsimulator/vehicles/parts/APartEngine.java
```diff
@@ -370,6 +373,9 @@ public void startEngine(){
 			WrapperAudio.playQuickSound(new SoundInstance(this, definition.packID + ":" + definition.systemName + "_starting"));
 			WrapperAudio.playQuickSound(new SoundInstance(this, definition.packID + ":" + definition.systemName + "_running", true));
 			WrapperAudio.playQuickSound(new SoundInstance(this, definition.packID + ":" + definition.systemName + "_supercharger", true));
+			WrapperAudio.playQuickSound(new SoundInstance(this, definition.engine.customSound1, true));
+			WrapperAudio.playQuickSound(new SoundInstance(this, definition.engine.customSound2, true));
+			WrapperAudio.playQuickSound(new SoundInstance(this, definition.engine.customSound3, true));
```
fsendventd, line 378: These three lines look really confusing, considering how the other engine sounds have been defined. Definition.engine.customsound1 isn't going to make any sense, because then people will be naming their sounds whatever. Do it like the others, but with more descriptive underscores like _idle, _rev, and _whatever-this-one's-for.

src/main/java/minecrafttransportsimulator/vehicles/parts/APartEngine.java
```diff
@@ -439,24 +445,63 @@ public void updateProviderSound(SoundInstance sound){
...
 			if(!state.running && internalFuel == 0){
 				sound.stop();
 			}else{
-				//Pitch should be 0.35 at idle, with a 0.35 increase for every 2500 RPM, or every 25000 RPM for jet (high-revving) engines.
-				sound.pitch = (float) (0.35*(1 + Math.max(0, (RPM - engineStartRPM))/(definition.engine.maxRPM < 15000 ? 500 : 5000)));
+				//Pitch should be 0.35 at idle, with a 0.35 increase for every 2500 RPM, or every 25000 RPM for jet (high-revving) engines by default.
+				//Custom pitched engines allow for more/less dynamic engine sounds without game crashing or a "one fix all" solution that fixes nothing
+				if (definition.engine.pitchRev > 0){
+						sound.pitch = (float) (((definition.engine.pitchRev*((1 + Math.max(0, ((RPM - engineStartRPM))/definition.engine.maxRPM))))- definition.engine.pitchRev)+ definition.engine.pitchIdle);
+						sound.volume = (float) (((definition.engine.volumeRev*((1 + Math.max(0, ((RPM - engineStartRPM))/definition.engine.maxRPM))))- definition.engine.volumeRev)+ definition.engine.volumeIdle);
```
fsendventd, line 465: Engines never had a volume change as far as I could tell. Did you get this out of the superchargers?

src/main/java/minecrafttransportsimulator/vehicles/parts/APartEngine.java
```diff
@@ -439,24 +445,63 @@ public void updateProviderSound(SoundInstance sound){
...
+				}
 			}
 		}else if(sound.soundName.endsWith("_running")){
 			if(!state.running && internalFuel == 0){
 				sound.stop();
 			}else{
-				//Pitch should be 0.35 at idle, with a 0.35 increase for every 2500 RPM, or every 25000 RPM for jet (high-revving) engines.
-				sound.pitch = (float) (0.35*(1 + Math.max(0, (RPM - engineStartRPM))/(definition.engine.maxRPM < 15000 ? 500 : 5000)));
+				//Pitch should be 0.35 at idle, with a 0.35 increase for every 2500 RPM, or every 25000 RPM for jet (high-revving) engines by default.
+				//Custom pitched engines allow for more/less dynamic engine sounds without game crashing or a "one fix all" solution that fixes nothing
```
fsendventd, line 462: Fix your grammar here.

src/main/java/minecrafttransportsimulator/vehicles/parts/APartEngineGeared.java
```diff
@@ -22,10 +22,18 @@ public void updatePart(){
 		//Do automatic transmission functions if needed.
 		if(state.running && definition.engine.isAutomatic){
 			if(currentGear > 0){
-				if(RPM > getSafeRPMFromMax(this.definition.engine.maxRPM)*0.5F*(1.0F + vehicle.throttle/100F)){
-					shiftUp(false);
-				}else if(RPM < getSafeRPMFromMax(this.definition.engine.maxRPM)*0.25*(1.0F + vehicle.throttle/100F) && currentGear > 1){
-					shiftDown(false);
+				if (!definition.engine.customShifter) {
+					if(RPM > getSafeRPMFromMax(this.definition.engine.maxRPM)*0.47F*(1.0F + vehicle.throttle/100F)){
```
fsendventd, line 26: 0.47F here used to be 0.5F. Put that back so nothing changes for people who didn't use custom shift points.

src/main/java/minecrafttransportsimulator/vehicles/parts/APartEngineGeared.java
```diff
@@ -76,7 +84,7 @@ public void shiftUp(boolean packet){
 		if(currentGear == -1){
 			currentGear = 0;
 		}else if(currentGear == 0){
-			if(vehicle.velocity > -0.1){
+			if(vehicle.velocity > -0.4){
```
fsendventd, line 87: Change 0.4 back to 0.1. You can't go making sneaky changes like that.

src/main/java/minecrafttransportsimulator/vehicles/parts/APartEngineGeared.java
```diff
@@ -98,7 +106,7 @@ public void shiftDown(boolean packet){
 				--currentGear;
 			}
 		}else if(currentGear == 0){
-			if(vehicle.velocity < 0.1){
+			if(vehicle.velocity < 0.4){
```
fsendventd, line 109: Same here.

src/main/java/minecrafttransportsimulator/vehicles/parts/APartEngineGeared.java
DonBruce64: I agree with this.  Don't want to change built-in code.  That's for me to mess with.  Yes, this makes the engine shift sooner, but there's a reason I have values the value they are.

src/main/java/minecrafttransportsimulator/vehicles/parts/APartEngine.java
DonBruce64: I get the lines, so no issue here.  But perhaps a more descriptive name is indeed in order?  For sounds, you do specify the packID in the JSON (to let you use other JSON's sounds), so it would be prudent to do the same here.  On the other hand, the *current* engine sounds are specific to that engine.  Not sure what to do here....

src/main/java/minecrafttransportsimulator/vehicles/parts/APartEngine.java
DonBruce64: I'm okay with it IF and only IF it doesn't affect current engines.  This code should only be active if the engine is on of the fancy types with pitch clamping or what-have-yous.  I can understand wanting multiple sounds and needing to shut one of them up to play another.

src/main/java/minecrafttransportsimulator/vehicles/parts/APartEngineGeared.java
DonBruce64: At a minimum make it an even decimal.  I'd prefer 0.25, as that's a speed of 5 blocks a second.  0.4 is a bit too fast IMO.  Granted, speedFactor applies, but that's not *supposed* to be accounted for at all in the physics code.  Just in the movement code.

PR #398, opened 2020-04-20
Realistic acceleration and braking animations for non-threaded/non-tracked ground vehicles

RishiMenon2004:
I've added 2 animation variables, "acceleration" and "braking", for realistic simulation of suspensions.
I'm hoping this would be integrated in the actual physics of the vehicles in future updates.

### DIFF
--- a/src/main/java/minecrafttransportsimulator/vehicles/main/EntityVehicleE_Powered.java
+++ b/src/main/java/minecrafttransportsimulator/vehicles/main/EntityVehicleE_Powered.java
@@ -49,7 +49,6 @@ public abstract class EntityVehicleE_Powered extends EntityVehicleD_Moving imple
 	public boolean hornOn;
 	public boolean sirenOn;
 	
-	public byte throttle;
 	public byte totalGuns = 0;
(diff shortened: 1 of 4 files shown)


src/main/java/minecrafttransportsimulator/vehicles/main/EntityVehicleE_Powered.java
```diff
@@ -389,8 +388,7 @@ public void readFromNBT(NBTTagCompound tagCompound){
     
 	@Override
 	public NBTTagCompound writeToNBT(NBTTagCompound tagCompound){
-		super.writeToNBT(tagCompound);		
-		tagCompound.setByte("throttle", this.throttle);
```
DonBruce64, line 393: You got the writeToNBT setByte saving, but you didn't get the readFromNBT getByte loading.  Gonna have to move that too.

RishiMenon2004: Ok, I've fixed it

PR #424, opened 2020-05-11
Fix ItemWrench again

SteveTheEngineer:


DonBruce64:
I see _someone_ didn't remember Java order of operations.  Given this and the last PR I'm starting to wonder if you're even testing these before submitting them...

SteveTheEngineer:
> I see _someone_ didn't remember Java order of operations. Given this and the last PR I'm starting to wonder if you're even testing these before submitting them...

I'm sorry, I've made the first commit in github web as I was too lazy to do it in IDEA and test it. However now I've tested it and it works as intended.

PR #425, opened 2020-05-12
Removed cutoff of acceleration and made sure that the animation was not called when the pbrake is on

RishiMenon2004:
I realised that the cutoff for the acceleration was janky, so I removed it in the code and made it so that the clamps could work as the cutoff. I also added the variables to the handbook in the ground vehicles section. But my libre office is a little broken, so I don't know if it affected the formatting. It did give caution that it might effect.


RishiMenon2004:
Oh, dang the formatting is broken

RishiMenon2004:
I'll fix the formatting and push another commit

RishiMenon2004:
oops

DonBruce64:
So formatting is all good now?

RishiMenon2004:
> So formatting is all good now?
yep, it's all good now
and I've fixed a certain flickering issue that appeared when the vehicles was slightly shifting when the pbrake isn't on

PR #448, opened 2020-06-04
Small adjustment to propeller pitch code, to allow for better performance on piston engines. 

fsendventd:
What it says on the tin. The way it was working before, they weren't as efficient as they could have been.


fsendventd:
> 
> 
> I see static values. Those are always bad as they don't handle multiple JSON values reliably. Please change.

I set them to static values to prevent the issues we had before with jet engines not spinning up high enough. A solid 100-150 RPM difference helps piston engines run better but won't break turboprops again.

DonBruce64:
Not merging as other PR had same changes.

PR #539, opened 2020-10-04
CustomSoundset "PitchAdvanced" & "VolumeAdvanced" fix

BlazeWolf1122:
BE GONE, **_J A N K ! ! !_**
Simply put, fixes the issue of audio cutting in and out, and replaces it with fading in and out, so the system is actually useful


DonBruce64:
Cancelling as these changes were merged via Commit: e93fbaa1ad346144143ca89b459d61c491a2c29d

PR #558, opened 2020-10-14
Semi "Realistic" car handling

BlazeWolf1122:
It's a bit clunky, but does the job.
I literally have no clue how to change it to be more optimized

From a moment to moment standpoint, cars shouldn't be too heavily effected by this change.  Most tires have a "lateral friction" of 0.5, so the driver would have to put on less grippier wheels, or try very hard to make these effects visible. Divided "groundDevice.getLateralFriction()" by 2, to make the effects more visible

As per usual with the PRs I make, i'd prefer communication is in the discord server rather than on github itself. it's kinda hard for me to access github sometimes

LOCAL FORD MUSTANG CRASHES INTO [REDACTED]

Subaru AWD donuts

UNDERSTEER

### DIFF
--- a/src/main/java/minecrafttransportsimulator/vehicles/main/EntityVehicleD_Moving.java
+++ b/src/main/java/minecrafttransportsimulator/vehicles/main/EntityVehicleD_Moving.java
@@ -56,15 +56,15 @@ abstract class EntityVehicleD_Moving extends EntityVehicleC_Colliding{
     	private double driftForce = 0;
   	private double driveTrain = 0;
   	{
-  		if (this.definition.motorized.isFrontWheelDrive && this.definition.motorized.isRearWheelDrive){
-  			driveTrain = 100;
-  		}else if (definition.motorized.isRearWheelDrive){
-  			driveTrain = 20;
-  		}else if (definition.motorized.isFrontWheelDrive){
-  			driveTrain = -20; 
-  		}else{
-  			driveTrain = 100;
-  		}
+  	if (this.definition.motorized.isFrontWheelDrive && this.definition.motorized.isRearWheelDrive){
+  		driveTrain = 100;
+  	}else if (definition.motorized.isRearWheelDrive){
+  		driveTrain = 20;
+  	}else if (definition.motorized.isFrontWheelDrive){
+  		driveTrain = -20; 
+  	}else{
+  		driveTrain = 100;
+  	}
   	}
 	
 	/**List of ground devices on the ground.  Populated after each movement to be used in turning/braking calculations.*/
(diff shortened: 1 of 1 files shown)


src/main/java/minecrafttransportsimulator/vehicles/main/EntityVehicleD_Moving.java
```diff
@@ -52,7 +53,20 @@
 	private final Point3d normalizedGroundVelocityVector = new Point3d(0, 0, 0);
 	private final Point3d normalizedGroundHeadingVector = new Point3d(0, 0, 0);
   	private final VehicleGroundDeviceCollection groundDeviceBoxes;
-    
+    	private double driftForce = 0;
+  	private double driveTrain = 0;
+  	{
+  	if (this.definition.motorized.isFrontWheelDrive == true && this.definition.motorized.isRearWheelDrive == true) {
```
DonBruce64, line 59: This makes me cry.  You *never* need to check if a boolean data type == true.  By definition, it's either "true" or "false".  You're basically asking: "does true == true".  Also, you don't need to prefix with "this".  "this" is implied.

src/main/java/minecrafttransportsimulator/vehicles/main/EntityVehicleD_Moving.java
```diff
@@ -234,6 +248,36 @@ private float getBrakingForce(){
 		return brakingFactor;
 	}
 	
+	/**
+	 * Returns rotating force for skidding, based on the vehicle drivetrain and other usual MTS calculations
+	 */
+	private double getDriveTrainSkiddingForce(){
+		//First check grounded ground devices.
+		for(PartGroundDevice groundDevice : groundedGroundDevices){
```
DonBruce64, line 256: Why are you checking all ground devices?  You're not doing anything with them....  Calculating all this math 4+ times a vehicle will eat into performance.  Cross product math especially.

src/main/java/minecrafttransportsimulator/vehicles/main/EntityVehicleD_Moving.java
```diff
@@ -234,6 +248,36 @@ private float getBrakingForce(){
...
+			if(goingInReverse && dotProduct < 0){
+				if(vectorDelta >= 90){
+					vectorDelta = -(180 - vectorDelta);
+				}else if(vectorDelta <= -90){
+					vectorDelta = 180 + vectorDelta;
+				}
+			}
+			
+				//Get factor of how much we can "correct" our turning.
+					driftForce = vectorDelta * (skiddingFactor / driveTrain);
```
DonBruce64, line 273: This isn't the right indent location.

DonBruce64 (changes requested):

Issue #543, opened 2020-10-06
[Suggestion] Potion effects for riders of a vehicle


gyrohero:
My intent with this is to allow certain vehicles to cause all of their riders to incur potion effects depending on the purpose of the vehicle. My inspiration was the idea of having military helicopters give their riders the night vision effect. Other possible applications would be an armored vehicle that gives its riders the resistance effect, or an ambulance/medevac helicopter that gives its riders a small amount of regeneration. Using the existing MC potion effect framework should limit how much work is needed for this, while yielding some nice extra possibilities for pack authors.

As for implementation, here's how I see it happening:

- Within the mcinterface package (possibly WrapperEntity), create some method like .addEffect that will call the .addPotionEffect() method on an EntityLivingBase representing the rider. I'm not sure exactly how to check that the rider is an EntityLivingBase and not just an Entity, as I'm still new to Java and MC mod development.
- Within EntityVehicleB_Rideable, call the .addEffect method on the rider within the addRider and updateRider methods, to ensure the effect is continuously applied.
- As for the JSONdef, I see the important properties being:
  1) the name of the effect (can be drawn from the names enumerated in net.minecraft.init.MobEffects)
  2) the amplifier
  3) the duration, in ticks -- this will let pack authors figure out how to balance the effect not wearing out while still in the vehicle vs. the effect going away soon after exiting. The amount of appropriate residual effects may vary with the application anyway.

If you think this seems feasible, I may be able to work on some code on my own. The biggest uncertainty for me is the Entity vs. EntityLivingBase thing I mentioned.

DonBruce64:
I'd say it's feasible for sure.  All the places you said where the hooks need to be are correct.  As to testing if the entity is an EntityLivingBase, you can just do "if(entity instanceof EntityLivingBase)".  Throw a PR and I'll take a look, and if it works, I'll put it in.  Just don't forget the handbook!

PR #566, opened 2020-10-20
Adds VehicleEffects and engine RPM % variables

gyrohero:
Addresses issues #543 (effects) and #567 (RPM %)


src/main/java/minecrafttransportsimulator/vehicles/main/EntityVehicleB_Rideable.java
```diff
@@ -109,11 +125,27 @@ public void removeRider(IWrapperEntity rider, Iterator<IWrapperEntity> iterator)
...
+		if(this.definition.effects != null) {
+			for(VehicleEffect effect: this.definition.effects) {
+				rider.removeEffect(effect.name);
+			}
+		}
+			
 		
 		//Set the rider dismount position if we are on the server.
 		//If we are on the client, disable mouse-yoke blocking.
 		if(!world.isClient()){
```
DonBruce64, line 139: So, I get why you put the code to get rid of the effects here.  Only concern is that this block only executes on the server.  Will that cause syncing issues?  Effects should be either applied and removed on server/client, or only the server, but never a mix of both or de-syncs can occur.

src/main/java/mcinterface1122/WrapperEntity.java
```diff
@@ -167,6 +169,33 @@ public Point3d getRenderedPosition(float partialTicks){
 	}
 	private final Point3d mutableRenderPosition = new Point3d(0D, 0D, 0D);
 	
+	@Override
+	public void addEffect(String potionEffectName, int durationIn, int amplifierIn) {
+		// Only instances of EntityLivingBase can receive potion effects
+		if((entity instanceof EntityLivingBase)) {
+			try{
```
DonBruce64, line 176: While I understand the idea of having a try/catch block here, a better way of doing this is PotionEffect = new PotionEffect(),    if(potionEffect == null) throw Exception.  Either that, or the Potion itself will be null.  Basically, something is assured to be null, so it's better to use a null check to throw than a try/catch.  Usually try/catch is for file I/O stuff like JSON parsing.

DonBruce64 (changes requested):
Couple of comments.  Also, somehow a WRL0001.tmp file found its way in here....

src/main/java/mcinterface1122/WrapperEntity.java
gyrohero: Ok, that makes sense. Fixed.

src/main/java/minecrafttransportsimulator/vehicles/main/EntityVehicleB_Rideable.java
gyrohero: Ah, that was an oversight on my part. I just put it there so it would be after the 
`VehiclePart packPart = getPackDefForLocation(riderLocation);`

That being said, would it cause any issues if I move that line out of the if-block that checks if we're on the server?

gyrohero: Assuming that doesn't cause issues, it should be fixed now.

DonBruce64: Not an issue at all.  I just had it there because I didn't want to create the linking on the server if I didn't need it.  But now you need it, so out it goes.

Issue #582, opened 2020-10-28
[Suggestion] Optional muzzlePosition parameter for gun parts


gyrohero:
This parameter, if used, would allow determining where bullets are spawned, independent from the gun's center.

Some vehicles feature hardpoints that can support different types of weapons. Supporting weapons of different size means that it is best for the point of attachment to be at 0,0,0. However, doing so causes bullets to be spawned there as well, which can be somewhat inaccurate.

(Bonus suggestion) Use a list of muzzlePositions to represent multiple barrels, rocket tubes, missile racks, bomb bay drop points, etc. If the number of muzzlePositions given matches the capacity of the gun, then use (capacity - ammo_count) as the index of muzzlePositions to use when the next bullet is fired. This way, rocket/missile visibility can be toggled off when a rocket/missile has been fired from its respective muzzlePosition.

If the number of muzzlePositions does not match the capacity of the gun, just iterate through the muzzlePositions each time a bullet is fired. This would be useful for multi-barreled weapons like anti-air guns, bomber turrets, and ship guns. If the list only includes one muzzlePosition, it would just use that position, just like the original functionality I mentioned.

gyrohero:
If you give me the go-ahead on this, I may try to work on it myself and submit a PR

LauraDarkez:
Having muzzleposition list would help dpl get his slug working both aesthetically and physically. As long as you list positions in order of firing sequence

DonBruce64:
I agree with all of this.  Just be aware that currently, the bullet in a gun fires at the center, offset by the barrelLength.  This can probably be supplanted by the muzzlePositions, as for most guns there's no drawback to spawning at the center of the gun except a little slower bullet movement.  Also, be aware that some guns' centers don't match up with their muzzle ends.  Like tank turrets, for example.  Those tilt upwards off-center.  Food for thought.

PR #584, opened 2020-10-29
Adds an optional list of muzzle positions to guns

gyrohero:
Addresses #582


src/main/java/minecrafttransportsimulator/vehicles/parts/PartGun.java
```diff
@@ -46,6 +48,7 @@ public PartGun(EntityVehicleF_Physics vehicle, VehiclePart packVehicleDef, ItemP
 		this.bulletsLeft = data.getInteger("bulletsLeft");
 		this.currentOrientation = data.getPoint3d("currentOrientation");
 		this.prevOrientation = currentOrientation.copy();
+		this.uniqueMuzzles = (definition.gun.muzzleObjects != null) && definition.gun.muzzleObjects.size() == definition.gun.capacity;
```
DonBruce64, line 51: While I understand storing this value, I wouldn't store it globally for the entire part instance.  This prevents the changes made in devMode from affecting the gun firing.  Instead, I'd put it as a boolean down in the getFiringPosition() routine.  Especially since down there you already check if the muzzleObjects is non-null to do custom muzzle position logic.

src/main/java/minecrafttransportsimulator/vehicles/parts/PartGun.java
```diff
@@ -397,4 +400,20 @@ public void spawnParticles(){
...
 		}
 	}
+	
+	public Point3d getFiringPosition() {
+		if (definition.gun.muzzleObjects != null) {
+			//If muzzle count is the same as capacity, use the muzzles in order
+			//Otherwise, iterate through the available muzzles
+			int muzzleIndex = this.uniqueMuzzles ? definition.gun.capacity - this.bulletsLeft : this.bulletsFired % definition.gun.muzzleObjects.size();
+			MuzzleObject currentMuzzle = definition.gun.muzzleObjects.get(muzzleIndex);
+			if(currentMuzzle != null) {
```
DonBruce64, line 410: The currentMuzzle != null check is redundant here.  You can't have a null entry in a list unless you have bad comma placement, and that's a whole different issue.  I personally prefer the game to crash when an error like this is encountered as it lets pack creators know they have an issue.  Otherwise if it just used the default, they might not notice the issue as the gun would still be firing.

DonBruce64 (changes requested):

src/main/java/minecrafttransportsimulator/vehicles/parts/PartGun.java
gyrohero: Gotcha. My thought process was to reduce the workload by only computing this once, but I hadn't thought about the possibility that it could change via devMode. Also, it's not like this line is that operationally expensive anyway.

src/main/java/minecrafttransportsimulator/vehicles/parts/PartGun.java
gyrohero: Ok, I'll take out that check then.

PR #592, opened 2020-11-05
Fixes muzzle position not mirroring with part

gyrohero:
@LauraDarkez caught this bug and tested the fix for me. The fix itself worked, but she said that it also made her discover a potential problem with the mirroring system


DonBruce64:
Going to have to decline this one.  The mirroring issue is more than just the muzzlePosition.  It also makes the gun turn the wrong way in general.  I've seen it happen frequently when you put the ragingSeagull gunner seat on cars and attach guns to them.  The proper way to fix such a thing, I think, is to flag the seat as disableMirroring : true to ensure that the rotated parts on it don't mirror themselves when they rotate.  Please let me know if that works prior to this, as I'd rather not hard-code mirrored gun logic as mirroring is designed to purely be a rendering thing only.  Though if this fix corrects both the animation and the bullet itself, that's okay.

LauraDarkez:
don i would like to point out that my lightning rocket launcher is meant to mirror on the biplane, as the firing pattern is inner to outer on both units. gyro fixed the issue of muzzlepositions not following requested mirroring. and i will disableMirroring on that gunner seat, but overall having muzzlepositions mirror when mirroring is requested, is a good thing. the issue i noticed was that muzzles were reversed in logic. so now, the muzzles do mirror properly, but the rotation-mirroring is an issue that will exist unless they are disablemirrored. that issue doesnt exist on non-rotating guns though.

src/main/java/minecrafttransportsimulator/vehicles/parts/PartGun.java
```diff
@@ -404,9 +404,12 @@ public void spawnParticles(){
 	public Point3d getFiringPosition() {
 		//If muzzle count is the same as capacity, use the muzzles in order
 		//Otherwise, iterate through the available muzzles
+		//Mirror the position if it needs to be mirrored
 		if (definition.gun.muzzlePositions != null) {
 			currentMuzzle = definition.gun.muzzlePositions.size() == definition.gun.capacity ? definition.gun.capacity - this.bulletsLeft : this.bulletsFired % definition.gun.muzzlePositions.size();
-			return definition.gun.muzzlePositions.get(currentMuzzle).copy();
+			Point3d firingPosition = definition.gun.muzzlePositions.get(currentMuzzle).copy();
```
DonBruce64, line 410: I see you're copying the muzzlePosition here, but then making yet another new Point3d object later to return.  If you're making a copy of the object, you should be modifying the copy.  Or, if you want to make a new object, you can simply not make a copy of the muzzlePosition.  Personally, I'd keep the copy(), but change the latter to if(morringStuffs)firingPosition.x *= -1.  Basically, get a copy of the point, and if you need to mirror it, invert the x coord, and then return the copied variable.  This will be important as this logic happens every tick, and you don't want to create multiple objects if you can help it.

DonBruce64 (changes requested):

DonBruce64:
That is a fair and accurate assessment for this feature.  Got just one issue with the execution, and then I can pull it in.

src/main/java/mcinterface1122/WrapperEntity.java
```diff
@@ -60,6 +60,13 @@ public void setRiding(AEntityBase entityToRide){
 			entity.dismountRidingEntity();
 		}
 	}
+	@Override
+	public AEntityBase getBaseEntity() {
+		if (entity != null && entity instanceof BuilderEntity) {
```
DonBruce64, line 65: The != null check is redundant here, as instanceof takes care of that.  So you can just do:
return entity instanceof BuilderEntity ? ((BuilderEntity) entity).entity : null;

src/main/java/minecrafttransportsimulator/rendering/instances/ParticleBullet.java
```diff
@@ -123,12 +127,22 @@ public void update(){
 		}
 		
 		//Now that we have checked for collision, adjust motion to compensate for bullet movement and gravity.
-		//Ignore this if the bullet has a (rocket motor) burnTime that hasn't yet expired
+		//Ignore this if the bullet has a (rocket motor) burnTime that hasn't yet expired,
+		//And if the bullet is still accelerating, increase the velocity appropriately.
 		if (this.burnTimeLeft == 0) {
+			MasterLoader.coreInterface.logError("Slowing down and falling");
```
DonBruce64, line 133: Pretty sure you don't want this in the logs.  Pro-tip though: if you're using Eclipse or IDEA, System.out.println() works just as well, and doesn't bloat up the logs.

src/main/java/minecrafttransportsimulator/rendering/instances/ParticleBullet.java
```diff
@@ -123,12 +127,22 @@ public void update(){
...
 			motion.multiply(0.98D);
 			motion.y -= 0.0245D;
 		}
+		else if(this.accelerationLeft > 0) {
+			MasterLoader.coreInterface.logError("Current velocity " + velocity + ", " + motion);
```
DonBruce64, line 138: Same as other comment about this not being here.  And for the following lines too.

src/main/java/minecrafttransportsimulator/rendering/instances/ParticleMissile.java
```diff
@@ -44,14 +48,45 @@ public void update() {
 		//Otherwise, use the last target position
 		//If there is no valid target position, just continue forward
 		if (entityTarget != null && entityTarget.isValid()) {
-			targetPosition = entityTarget.getPosition();
+			//If the target is a vehicle, and we're not currently locked onto an engine,
+			//Try to find a valid, warm engine to track.
+			//If we were tracking an engine, but it is no longer valid or is cold, find a new one.
+			AEntityBase entityTargetBase = entityTarget.getBaseEntity();
+			if(entityTargetBase != null && entityTargetBase instanceof EntityVehicleF_Physics && (engineTarget == null || !engineTarget.isValid || engineTarget.temp < 30f)) {
```
DonBruce64, line 55: Don't need a null check if you're using instanceof.

src/main/java/minecrafttransportsimulator/rendering/instances/ParticleMissile.java
```diff
@@ -44,14 +48,45 @@ public void update() {
...
+						float distanceToEngine = (float)position.distanceTo(currentEngine.worldPos);
+						if (nearestEngine == null || distanceToEngine < smallestDistance) {
+							nearestEngine = currentEngine;
+							smallestDistance = distanceToEngine;
+						}
+					}
+				}
+				engineTarget = nearestEngine != null ? nearestEngine : null;
+			}
+			//If there isn't a valid engine, just track the entity
```
DonBruce64, line 75: Is this valid though?  Should a heat-seeking missle be tracking not-hot vehicles?  Pretty sure it shouldn't know the difference between a cold tank that isn't running and a building, so if the engine is cold because it's turned off, then it shouldn't try to track it.  Just my two pennies.

src/main/java/minecrafttransportsimulator/packets/instances/PacketBulletHit.java
```diff
@@ -80,6 +80,7 @@ public boolean handle(IWrapperWorld world, IWrapperPlayer player, EntityVehicleF
 			//Get the bullet definition, and the position the bullet hit.  Also get the gun that fired the bullet.
 			//We need this to make sure that this isn't a duplicate packet from another client.
 			JSONPart bulletDefinition = bullet.definition;
+			if (bulletDefinition.bullet.blastStrength == 0f) bulletDefinition.bullet.blastStrength = 1f;
```
DonBruce64, line 83: Is this for bomb strength?  Like TNT?  If so, I'd like to talk about this parameter, as I think this might be the wrong way to handle such a thing.

DonBruce64 (changes requested):

src/main/java/minecrafttransportsimulator/rendering/instances/ParticleMissile.java
gyrohero: Fair enough

PR #599, opened 2020-11-13
Fixes rockets sometimes firing in the wrong direction

gyrohero:
Addresses #597 

The problem was essentially that if a rocket (bullet with an accelerationTime defined) was given a low enough muzzleVelocity, then vehicle motion would comprise a significant amount of its overall motion. The bugged system was applying acceleration by just multiplying the bullet motion, in some cases accelerating it backwards or sideways relative to the direction the gun was facing.

My fix was to store a unit vector of the gun's direction at the time of firing and pass this along in the bullet constructor. Then when accelerating, it would add this vector multiplied by the amount of acceleration desired. But it uses the previous calculations for missiles since they need to be able to accelerate in directions other than the initial one.

### DIFF
--- a/src/main/java/minecrafttransportsimulator/rendering/instances/ParticleBullet.java
+++ b/src/main/java/minecrafttransportsimulator/rendering/instances/ParticleBullet.java
@@ -40,6 +40,7 @@ public class ParticleBullet extends AParticle{
 	private final PartGun gun;
 	private final int bulletNumber;
 	private final double initialVelocity;
+	private final Point3d initialDirection;
 	private final double deltaVelocity;
 	private final IWrapperEntity gunController;
 	private final Point3d blockMotionStep = new Point3d(0D, 0D, 0D);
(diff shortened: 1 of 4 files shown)


gyrohero:
I know I can merge this pull request myself now, but I just want to make sure I didn't do a bad with adding a parameter to the bullet constructor or something

src/main/java/minecrafttransportsimulator/rendering/instances/ParticleBullet.java
```diff
@@ -62,7 +63,7 @@ public ParticleBullet(Point3d position, Point3d motion, ItemPart bullet, PartGun
         this.deltaVelocity = accelerationLeft != 0 ? (bullet.definition.bullet.maxVelocity/20D/10D - initialVelocity) / accelerationLeft : 0D;
         this.gunController = gunController;
         this.timeUntilAirBurst = bullet.definition.bullet.airBurstDelay;
-        if(bullet.definition.bullet.blastStrength == 0F) bullet.definition.bullet.blastStrength = 1F;
```
DonBruce64, line 65: Did you mean to remove blastStrength here?

DonBruce64:
Just a single comment on something you may not have meant.  BTW, when you merge this PR, don't close the ticket.  The idea is to leave tickets open until the version drops on Curse with the fix, in case someone else has the same issue.

src/main/java/minecrafttransportsimulator/rendering/instances/ParticleBullet.java
gyrohero: Yep. This was leftover from the previous blastStrength implementation. With this line, any explosive bullets that aren't using blastStrength have a strength of one TNT

gyrohero:
Got it. Will do

Issue #602, opened 2020-11-13
[Enhancement] Variables for seat occupied and rider yaw and pitch


gyrohero:
In the past, we've discussed ways for a camera to track a player's sight without needing to have a gun. My proposed solution: have variables player_yaw_# and player_pitch_#. This would allow the variables to be used for other things like spotlights as well. The # would refer to the seat number in which to find the player. If the camera/light/etc. is to be controllable from multiple seats, have multiple animations, one for each seat.

Now, how do we handle possibly having players in multiple seats? Allowing the animations system to use && and ||, plus their NOT variants. These could be nested for more complex logic.
`"and": {
"variable": "customVariable1",
"nand": {"variable": "customVariable2"}
}`
At least initially, these could just accept boolean variables. That would allow having a switch to change control between 2 pilots. Or, with an added "player_in_seat_#" variable, you can do something like
`"nand": {"variable": "player_in_seat_1"}
`
to assign priority over which seat gets control. In this case, seat 2 would only get control if seat 1 was not occupied by a player.

So to sum up, this is really 3 suggestions:

- [x] player_yaw and player_pitch variables
- [x] and/or/nand/nor
- [x] player_in_seat variable

And each of these, individually, don't seem like they should be too hard to pull off. So I can knock these out as long as you think it's feasible/a good idea.

DonBruce64:
Let me think a bit about this one.  The implementation of variables themselves is easy, but I need to think about how they'll work with the other systems and how the rendering pipeline will behave.

gyrohero:
Alright, no problem

DonBruce64:
So, I'm still mulling on the booleans, but here's what I think should be done with the players:
Add a new variable called seat_X, where you can have the following:
seat_occupied
seat_rider_yaw
seat_rider_pitch
That would work nicely with the existing hierarchical system, where if you have a seat that's a gun's controller by the gun being a sub-part of the seat, or the gun being an additional part of the seat, then you don't need to worry about numbers.  The seat variables will only return values if a player is in that controlling seat.  For the inverse, where a seat is on the gun, then you'd have to specify a _1 for the seat number, but that should be easy to handle as you'll already know the seat order.  This won't require much anything in the VAS either, as it'll be using an existing system.

DonBruce64:
Might be a good idea to split these tickets.  I think the booleans will be far tricker than the seat variables.

gyrohero:
Sounds good to me!

PR #604, opened 2020-11-14
New seat variables for occupied, rider yaw and pitch + text object-friendly heading variable

gyrohero:
Implements #602 

This worked in testing for both cameras and object animation. The only limitation I noticed is that when using clamps, the player can continue to turn past the clamp and the camera will just stop turning until the player reaches the opposite clamp. But I don't see that as a _huge_ problem personally. And it seems like it could be problematic/difficult to limit the player's view, but only when the camera is active. Let me know what you think about it.

I also threw in an unrelated but useful "heading" variable that's text object-friendly for use in my Primary Flight Display.


gyrohero:
Hang on... on further testing there seems to be some bad math with pitch, but only for object animations, not cameras...

DonBruce64:
What's with the heading variable?  It's that the same as yaw, or the angle the vehicle is heading?

PR #605, opened 2020-11-15
Adds inhibitors, fixes fake ground device collision boxes, and other small fixes

gyrohero:
Mostly looking for you to review the inhibitor code to see if I did anything you don't like there


src/main/java/minecrafttransportsimulator/rendering/components/TransformInhibitor.java
```diff
@@ -0,0 +1,30 @@
...
+public class TransformInhibitor extends ATransformRenderable{
+	
+	public TransformInhibitor(VehicleAnimationDefinition definition){
+		super(definition);
+	}
+	
+	@Override
+	public boolean shouldInhibit(EntityVehicleF_Physics vehicle, APart optionalPart, float partialTicks) {
+		double value = definition.offset + getClock(vehicle).getFactoredState(vehicle, VehicleAnimationSystem.getVariableValue(definition.variable,  partialTicks, vehicle, optionalPart));
+		return value >= definition.clampMin && value <= definition.clampMax;
```
DonBruce64, line 23: In the file comment, you say this disables the animations if the value is not within the min and max clamps.  But from what I can tell, it disables the animation if it IS within the clamps.  I think your comment is backwards?  Or I just have Bad Maths.

DonBruce64:
Code looks just dandy.  Comment on comment is all I have.

src/main/java/minecrafttransportsimulator/rendering/components/TransformInhibitor.java
gyrohero: You're right, the comment is backwards

Issue #607, opened 2020-11-16
[Enhancement] Variables for missile incoming


gyrohero:
This would allow for instruments that provide a warning if the vehicle is being tracked by a missile. There would be 3 variables: missile_incoming (Boolean), missile_direction (float), and missile_distance (float). If there are multiple missiles incoming, this could be limited to whichever is closest, or possibly be numbered to allow a certain number of alerts.

For implementation, the missile would just have to pass a pointer of itself to the vehicle it's locker onto, simulating its laser or radar signature.

DonBruce64:
Should be simple.  And would be good as you could link it to an animation for an in-cockpit warning sound.  When I get to overhauling the instruments to work with the VAS, then the instrument can play the sound, so it'll be cross-compatible with different vehicles.

gyrohero:
I was thinking that too about the sound thing. Alright, cool.

fsendventd:
_WOMP WOMP WOMP_
Would this theoretically allow chaff or flares to be automatically deployed as well, once that's added?

gyrohero:
@fsendventd Yeah, we can probably set it up so that they deploy based on JSON-defined clamps of the distance/direction variables

DonBruce64:
I hesitate to have flares automatically deploy.  If only because there's not really a point in having missiles that home and lock on if they just automatically get chaffed. Plus, having chaff be part of the VAS is kinda icky.  Unless you mean a parameter, but even then, if you have a lock-on variable then that can correlate to chaff deployment without messing with variables.

fsendventd:
@DonBruce64 You would eventually run out of flares, they'd have ammo like guns. Plenty of other games (e.g. War Thunder) have automatic flare deployment in certain vehicles, it isn't overpowered.

PR #612, opened 2020-11-19
Missile Incoming Variables

gyrohero:
Addresses #607


src/main/java/minecrafttransportsimulator/rendering/instances/ParticleBullet.java
```diff
@@ -93,6 +93,7 @@ public void update(){
 			
 			//If we hit something, don't process anything further.
 			if(age == maxAge){
+				isValid = false;
```
DonBruce64, line 96: We don't need this now as this is checked in the main particle code.  We should probably be checking isValid here instead, and remove the age = maxAge.

src/main/java/mcinterface1122/WrapperWorld.java
```diff
@@ -200,7 +200,7 @@ public WrapperEntity getEntityLookingAt(IWrapperEntity entityLooking, float sear
 		Vec3d losVector = new Vec3d(lookerLos.x, lookerLos.y, lookerLos.z);
 		for(Entity entity : world.getEntitiesWithinAABBExcludingEntity(mcLooker, mcLooker.getEntityBoundingBox().grow(searchRadius))){
 			float distance = mcLooker.getDistance(entity);
-			if(distance < 10f) continue;
+			if(entity.equals(mcLooker.getRidingEntity())) continue;
```
DonBruce64, line 203: Usually I don't see loops structured like this.  I think continue might have some overhead?  Most of the time it's "if condition, do things".  Not "if not condition go to next thing."  Plus, it'd be better to check if the looker is riding the selected entity before we calculate the distance as that's math that doesn't need to happen if we're riding the entity.

DonBruce64 (changes requested):
Couple of tweaks with respect to order of operations, but nothing major.

src/main/java/mcinterface1122/WrapperWorld.java
gyrohero: The distance calculation there was leftover from using the distance check to prevent firing missiles at yourself. I'll fix that and the continue thing

PR #623, opened 2020-11-24
Particle object overhaul

gyrohero:
Moves exhaust object functionality to the particle object system


src/main/java/minecrafttransportsimulator/jsondefs/JSONVehicle.java
```diff
@@ -151,14 +151,33 @@
...
+    		public String type;
+    		public String color;
+    		public String toColor;
+    		public float transparency;
+    		public float toTransparency;
+    		public float scale;
+    		public float toScale;
+        	public Point3d pos;
+        	@Deprecated
+        	public float velocity;
```
DonBruce64, line 176: What's this guy about?  Is there a reason you have two velocities and one that is new is now depreciated?

src/main/java/minecrafttransportsimulator/packloading/LegacyCompatSystem.java
```diff
@@ -237,6 +262,13 @@ private static void performPartLegacyCompats(JSONPart definition){
 				definition.bullet.types = new ArrayList<String>();
 				definition.bullet.types.add(definition.bullet.type);
 			}
+			if (definition.bullet.particleObjects != null) {
```
DonBruce64, line 265: Is there a reason for this check?  It looks like for all the legacy compats you set the velocityVector, so is there every a chance it would be null here.

src/main/java/minecrafttransportsimulator/jsondefs/JSONVehicle.java
gyrohero: The velocity I had been using for particle objects was just a float representing the -Z value, since I was only concerned with smoke trails and such. Now that we're replacing exhaust objects with them, it needs to be a whole Point3d. And since we can't have a float velocity and a Point3d velocity, I had to name the latter something different

src/main/java/minecrafttransportsimulator/packloading/LegacyCompatSystem.java
gyrohero: Yes. The exhaust compats are one thing, but this compat is for previous implementations of particle object, which used the float velocity I mentioned in the previous comment

PR #642, opened 2020-11-30
Default positions for guns

gyrohero:


build.gradle
```diff
@@ -11,6 +11,7 @@ buildscript {
 apply plugin: 'net.minecraftforge.gradle.forge'
 apply plugin: 'com.github.johnrengelman.shadow'
 
+
```
DonBruce64, line 14: I think this will cause the buildscript.sh to fault out if we have it here.  It cares about the line number here.

gyrohero: Whoops, I guess I did that when I was resolving a conflict due to the version number

PR #648, opened 2020-12-07
Added inverted variables

DrprofLuigi:
Added an inverted variable option for boolean variables for visible animations.


DonBruce64:
Going to have to reject this, as variable visibility can already be inverted with different clamps.

PR #660, opened 2020-12-24
Add day_time variable

crazygmr101:


DonBruce64:
Closing, but added the code on my end. I realized that "time" was a poor choice for what really was "tick" on the current time variable.  I've changed that, and now "time" is the actual world time like what you have here.

PR #681, opened 2021-01-19
added JSONDescription annotations in JSONIntrument.java

RishiMenon2004:


RishiMenon2004:
:/ github didn't let me save it and pushed me directly to a PR, so that's why you see this

PR #773, opened 2021-04-03
Oversteering/Understeering variable, "overSteer"

BlazeWolf1122:
A proper, cleaner version of the Pull Request i made months back for semi proper drift physics
It ain't the best, but it's good enough for me with the little skill i have
A little bit of jank exists sometimes at low speed, but it's an optional parameter so it shouldn't effect anything without this parameter


BlazeWolf1122:
~~also, no more "gdaskjdjdajadsjkdd" commits~~
This could be phased out by a vehicle thrust forces system that i made a ticket for before, however i don't see the viability of making a whole new system right now, like how the customSound set was phased out after almost a year later by the new sound system

BlazeWolf1122:
Pull Request closed at the moment because of unexpected events while testing
aka "back to the drawing board"

https://user-images.githubusercontent.com/63666200/113491965-70de9d80-9489-11eb-8f96-3198ada45372.mp4

PR #786, opened 2021-04-14
DownForce value change + overSteer addition

BlazeWolf1122:
almost all (if not, then all) the bugs of the overSteer PR were fixed, and a few changes were added
downForce code was changed (by DB himself) to be more reasonable, though requires a lot of vehicle JSON changes
it is most likely ideal to change the downForce value to about 10% to 20% of it's original value, which does make the oversteer value look a bit wonky right next to it


BlazeWolf1122:
Added a little whats it called thing for documentation, but i'm terrible at documentation
~~Should this need documentation~~

PR #810, opened 2021-05-06
downForce and steering "hotfix"

BlazeWolf1122:
cc found some issues with SF and steering physics
basically this nerfs steering force the higher the SF
not the best method in the world, but a car spinning in drunk circles at high speed isn't very nice of a fate either

### DIFF
--- a/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
+++ b/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
@@ -427,7 +427,7 @@ abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding{
 				double turningForce = steeringAngle/turningDistance;
 				//Decrease force by the speed of the vehicle.  If we are going fast, we can't turn as quickly.
 				if(groundVelocity > 0.35D){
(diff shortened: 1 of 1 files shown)


src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
```diff
@@ -427,7 +427,7 @@ private double getTurningForce(){
 				double turningForce = steeringAngle/turningDistance;
 				//Decrease force by the speed of the vehicle.  If we are going fast, we can't turn as quickly.
 				if(groundVelocity > 0.35D){
-					turningForce *= Math.pow(0.3F, (groundVelocity*(1 - definition.motorized.downForce) - 0.35D));
+					turningForce *= Math.pow(0.3F, (groundVelocity*(1 - definition.motorized.downForce) - 0.35D)) / (SPEED_FACTOR/0.5);
```
DonBruce64, line 430: Going to request you double-check that the transition from `groundVelocity < 0.35` to `groundVelocity > 0.35` isn't a "jump", because I suspect it is.  Also double-check maths to make sure higher values don't do crazy things.

DonBruce64 (changes requested):

BlazeWolf1122:
~~help i can't math~~

Issue #843, opened 2021-06-10
[PR Idea] Different interpolation types for animation


RishiMenon2004:
Basically if you have a duration, another parameter can be written which specifies the interpolation type. So stuff like ease in and ease out can be done in just one animation block instead of combining 2 or 3 blocks. Also, is a PR idea not request.

DonBruce64:
Going to close this as PR ideas should spawn discussion or PRs, not issue tickets.  Will address the PRs as they roll in.

PR #847, opened 2021-06-12
Different interpolation types for animation

RishiMenon2004:
PR from #843
Adds different easing types for boolean animations.
I'll push the wiki after the PR is merged

### DIFF
--- a/src/main/java/minecrafttransportsimulator/jsondefs/JSONAnimationDefinition.java
+++ b/src/main/java/minecrafttransportsimulator/jsondefs/JSONAnimationDefinition.java
@@ -87,6 +87,9 @@ public class JSONAnimationDefinition{
 	}
 	
 	public static enum AnimationEasingType{
+		@JSONDescription("Constant velocity")
+		LINEAR,
+		
 		@JSONDescription("Accelerating from zero velocity")
 		EASEINQUAD,
 		
--- a/src/main/java/minecrafttransportsimulator/jsondefs/JSONAnimationDefinition.java
+++ b/src/main/java/minecrafttransportsimulator/jsondefs/JSONAnimationDefinition.java
@@ -41,6 +41,9 @@ public class JSONAnimationDefinition{
 	@JSONDescription("The duration of this animation, in ticks.  Causes the animation to be interpolated over the duration for smooth movement.  Useful for things like doors.  Only works for variables that go from 0-1.")
 	public int duration;
 	
+	@JSONDescription("The type of easing interpolation to apply on the animation.")
+	public AnimationEasingType easing;
+	
 	@JSONDescription("How long this animation waits to start after the variable goes to 1, in ticks.  For example, an animation with a forwardsDelay of 20 and a duration of 40 would wait one second before moving, and then take two seconds to move its entire animation.")
 	public int forwardsDelay;
 	
@@ -78,5 +81,13 @@ public class JSONAnimationDefinition{
 		INHIBITOR,
 		@JSONDescription("Like Inhibitor, but in this case this animation will be resumed if it was inhibited.  There is no change if the animation wasn't inhibited to begin with.")
 		ACTIVATOR;
+	}
+	
+	public static enum AnimationEasingType{
+		EASEINQUAD,
+		
+		EASEOUTQUAD,
+		
+		EASEINOUTQUAD
 	}
 }
--- a/src/main/java/minecrafttransportsimulator/jsondefs/JSONAnimationDefinition.java
+++ b/src/main/java/minecrafttransportsimulator/jsondefs/JSONAnimationDefinition.java
@@ -89,76 +89,52 @@ public class JSONAnimationDefinition{
 	public static enum AnimationEasingType{
 		@JSONDescription("Constant velocity.")
 		LINEAR,
-		
 		@JSONDescription("Accelerating from zero velocity. Qudratic.")
 		EASEINQUAD,
-		
 		@JSONDescription("Decelerating to zero velocity. Qudratic.")
 		EASEOUTQUAD,
-		
 		@JSONDescription("Acceleration until halfway, then deceleration. Qudratic.")
 		EASEINOUTQUAD,
-		
 		@JSONDescription("Accelerating from zero velocity. Cubic.")
 		EASEINCUBIC,
-		
 		@JSONDescription("Decelerating to zero velocity. Cubic.")
 		EASEOUTCUBIC,
-		
 		@JSONDescription("Acceleration until halfway, then deceleration. Cubic.")
 		EASEINOUTCUBIC,
-		
 		@JSONDescription("Accelerating from zero velocity. Qudratic.")
 		EASEINQUART,
-		
 		@JSONDescription("Decelerating to zero velocity. Qudratic.")
 		EASEOUTQUART,
-		
 		@JSONDescription("Acceleration until halfway, then deceleration. Qudratic.")
 		EASEINOUTQUART,
-		
 		@JSONDescription("Accelerating from zero velocity. Quintic.")
 		EASEINQUINT,
-		
 		@JSONDescription("Decelerating to zero velocity. Quintic.")
 		EASEOUTQUINT,
-		
 		@JSONDescription("Acceleration until halfway, then deceleration. Quintic.")
 		EASEINOUTQUINT,
-		
 		@JSONDescription("Accelerating from zero velocity. Circular.")
 		EASEINCIRC,
-		
 		@JSONDescription("Decelerating to zero velocity. Circular.")
 		EASEOUTCIRC,
-		
 		@JSONDescription("Acceleration until halfway, then deceleration. Circular.")
 		EASEINOUTCIRC,
-		
 		@JSONDescription("Pulls back the velocity below 0 and then accelerates.")
 		EASEINBACK,
-		
 		@JSONDescription("Accelerates past 1 and then pulls back the velocity down to 1.")
 		EASEOUTBACK,
-		
 		@JSONDescription("Pulls back the velocity below 0. then accelerates past 1 and then pulls back the velocity down to 1.")
 		EASEINOUTBACK,
-		
 		@JSONDescription("Eases with an elastic effect at the beginning.")
 		EASEINELASTIC,
-		
 		@JSONDescription("Eases with an elastic effect at the end.")
 		EASEOUTELASTIC,
-		
 		@JSONDescription("Eases with an elastic effect at both the beginning and end.")
 		EASEINOUTELASTIC,
-		
 		@JSONDescription("Provides a bouncing animation at the beginning.")
 		EASEINBOUNCE,
-		
 		@JSONDescription("Provides a bouncing animation at the end.")
 		EASEOUTBOUNCE,
-		
 		@JSONDescription("Provides a bouncing animation at both the beginning and end.")
 		EASEINOUTBOUNCE
 	}
--- a/src/main/java/minecrafttransportsimulator/jsondefs/JSONAnimationDefinition.java
+++ b/src/main/java/minecrafttransportsimulator/jsondefs/JSONAnimationDefinition.java
@@ -87,16 +87,79 @@ public class JSONAnimationDefinition{
 	}
 	
 	public static enum AnimationEasingType{
-		@JSONDescription("Constant velocity")
+		@JSONDescription("Constant velocity.")
 		LINEAR,
 		
-		@JSONDescription("Accelerating from zero velocity")
+		@JSONDescription("Accelerating from zero velocity. Qudratic.")
 		EASEINQUAD,
 		
-		@JSONDescription("Decelerating to zero velocity")
+		@JSONDescription("Decelerating to zero velocity. Qudratic.")
 		EASEOUTQUAD,
 		
-		@JSONDescription("Acceleration until halfway, then deceleration")
-		EASEINOUTQUAD
+		@JSONDescription("Acceleration until halfway, then deceleration. Qudratic.")
+		EASEINOUTQUAD,
+		
+		@JSONDescription("Accelerating from zero velocity. Cubic.")
+		EASEINCUBIC,
+		
+		@JSONDescription("Decelerating to zero velocity. Cubic.")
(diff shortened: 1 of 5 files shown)


src/main/java/minecrafttransportsimulator/jsondefs/JSONAnimationDefinition.java
```diff
@@ -79,4 +85,81 @@
 		@JSONDescription("Like Inhibitor, but in this case this animation will be resumed if it was inhibited.  There is no change if the animation wasn't inhibited to begin with.")
 		ACTIVATOR;
 	}
+
```
DonBruce64, line 88: Normally I don't like extra spaces, but if the other enums in the other JSON files have em, them they're fine.  Just want uniformity over anything else.

src/main/java/minecrafttransportsimulator/rendering/components/AnimationVariableEasing.java
```diff
@@ -0,0 +1,512 @@
...
+	private static final double c3 = c1 + 1;
+	private static final double c4 = (2 * Math.PI) / 3;
+	private static final double c5 = (2 * Math.PI) / 4.5;
+	private static final double n1 = 7.5625;
+	private static final double d1 = 2.75;
+	
+	/* This is used to return the interpolated animation values
+	 * It call the interpolation methods equivalent to the easing type
+	 */
+	public double getEasingValue(JSONAnimationDefinition animation, long timeMoved, boolean isReverse) {
```
DonBruce64, line 25: Now here, spaces are no needed.  Makes the code a bit more "exploded" and a bit harder to follow on some screens.

src/main/java/minecrafttransportsimulator/rendering/components/AnimationVariableEasing.java
```diff
@@ -0,0 +1,512 @@
...
+				return time;
+			}
+		}
+	}
+	
+	/* This is used to check the easing type defined in the JSON
+	 * And call the respective easing function to return a value
+	 */
+	public double getEasingType(JSONAnimationDefinition.AnimationEasingType direction, double time) {
+		switch(direction) {
```
DonBruce64, line 66: Would prefer not calling a function here, and instead returning the value directly, given how simple some of the functions are.  Same comment on spacing.

src/main/java/minecrafttransportsimulator/rendering/components/AnimationVariableEasing.java
```diff
@@ -0,0 +1,512 @@
...
+		}
+	}
+	
+	/* Below are functions that calculate the graph of each easing curve
+	 * It returns a value according to the time elapsed in an animation
+	 */
+
+	public double easeInQuad(double time) {
+		
+		double value = time * time;
```
DonBruce64, line 181: Wha?  you don't need to create a new double variable.  Just return the value your'e assigning to the variable.
So: `return time*time;`

src/main/java/minecrafttransportsimulator/rendering/components/AnimationVariableEasing.java
```diff
@@ -0,0 +1,512 @@
...
+
+		return value;
+	}
+	
+	public double easeOutQuad(double time) {
+
+		double value = time * (2 - time);
```
DonBruce64, line 188: This is a one-line thing, so really doesn't need to be in a function.

src/main/java/minecrafttransportsimulator/rendering/components/AnimationVariableEasing.java
```diff
@@ -0,0 +1,512 @@
...
+		}
+	}
+	
+	/* This is used to check the easing type defined in the JSON
+	 * And call the respective easing function to return a value
+	 */
+	public double getEasingType(JSONAnimationDefinition.AnimationEasingType direction, double time) {
+		switch(direction) {
+		
+			case LINEAR: {
```
DonBruce64, line 68: FYI, switches that are only a single line can be written without brackets.
`case LINEAR: return time;`
It's more compact, and if you can consolidate all the returns to single lines, it's far easier to see the flow and list of enums in a single screen.  See the animation switches for examples of how this works.

src/main/java/minecrafttransportsimulator/rendering/components/AnimationVariableEasing.java
```diff
@@ -0,0 +1,512 @@
...
+	public double easeInQuart(double time) {
+		
+		double value = time * time * time * time;
+		
+		return value;
+	}
+	
+	public double easeOutQuart(double time) {
+		
+		double value = 1 - (--time) * time * time * time;
```
DonBruce64, line 250: I don't think you meant to do `--time` here?  That would make the variable time go down by 1.  Perhaps you mean minus `-time` instead?

src/main/java/minecrafttransportsimulator/rendering/components/DurationDelayClock.java
```diff
@@ -90,13 +92,16 @@ public double getFactoredState(AEntityC_Definable<?> entity, double value){
 		}
 		
 		double movementFactor = 0;
+		AnimationVariableEasing easing = new AnimationVariableEasing();
```
DonBruce64, line 95: No.  Please no.  Don't create a new easing object every check of the variable.  Rather, declare the methods in that object as static, and call the method statically via `AnimationVariableEasing.getEasingValue()`.  Those methods don't change based on anything.  It's all in the function.

Heck, as consolidated as it is, why not move that method into this class and make it static instead?  Then the code is right where you use it rather than in another class file that only does math.

DonBruce64 (changes requested):
The idea is there, but needs some cleanup.  Some issues are formatting, others are performance.

src/main/java/minecrafttransportsimulator/rendering/components/AnimationVariableEasing.java
RishiMenon2004: It's just to make it look uniform, some are really big, so i put them in a function. And having some in the switch case and some in functions seems nonuniform to me

src/main/java/minecrafttransportsimulator/rendering/components/AnimationVariableEasing.java
RishiMenon2004: 👌 done

src/main/java/minecrafttransportsimulator/rendering/components/AnimationVariableEasing.java
RishiMenon2004: fixed that

src/main/java/minecrafttransportsimulator/rendering/components/AnimationVariableEasing.java
RishiMenon2004: again, uniformity of the functions. :harold: please no angy

src/main/java/minecrafttransportsimulator/rendering/components/AnimationVariableEasing.java
RishiMenon2004: https://gist.github.com/gre/1650294#file-easing-js-L23
i don't think so?

RishiMenon2004: from what i see, it's ``1 - (t-1) * (t-1) * (t-1) * (t-1)``
and ``--t`` uses ``t-1`` for all the 4 ``t``s there

src/main/java/minecrafttransportsimulator/rendering/components/DurationDelayClock.java
RishiMenon2004: Hmm, ok, i initially moved it to the class cause the switch case wall still in the getEasingValue method. But the switch case is moved out, so i'll move it back

src/main/java/minecrafttransportsimulator/rendering/components/AnimationVariableEasing.java
DonBruce64: Ah, gotcha.  That makes sense then.  And is quite elegant, if I do say so!

src/main/java/minecrafttransportsimulator/rendering/components/DurationDelayClock.java
```diff
@@ -29,6 +31,43 @@ public DurationDelayClock(JSONAnimationDefinition animation){
 		this.isUseful = shouldDoFactoring || animation.animationType.equals(AnimationComponentType.VISIBILITY)  || animation.animationType.equals(AnimationComponentType.INHIBITOR)  || animation.animationType.equals(AnimationComponentType.ACTIVATOR) || animation.forwardsStartSound != null || animation.forwardsEndSound != null || animation.reverseStartSound != null || animation.reverseEndSound != null;
 	}
 	
+	/**Returns the interpolated animation values
+	 * It calls the interpolation methods equivalent to the easing type
+	 */
+	public double getEasingValue(JSONAnimationDefinition animation, long timeMoved, boolean isReverse) {
+		double time = timeMoved/(double)(animation.duration*50);
```
DonBruce64, line 38: So, looking at the code, isn't this movementFactor in the calling function?  If so, there's no reason to re-calculate it.

src/main/java/minecrafttransportsimulator/rendering/components/DurationDelayClock.java
```diff
@@ -29,6 +31,43 @@ public DurationDelayClock(JSONAnimationDefinition animation){
...
+		//Check if the animation is playing in reverse
+		if (isReverse) {
+			
+			//Check if reverseEasing is not omitted
+			if (animation.reverseEasing != null) {
+				return AnimationVariableEasing.getEasingType(animation.reverseEasing, time);
+				
+			//If it is, then check if forwardsEasing isn't omitted
+			//We can use it's value to apply an easing type to the reverse animation
+			} else if (animation.forwardsEasing != null) {
```
DonBruce64, line 49: Why does forwards easing apply in reverse?  But reverseEasing not apply in forwards?  I would think they'd be the same.  Is this a legacy thing, or just allowing for the same easing definition for both Forwards/Reverse in one area?  If so, I don't see it needing to be here as I can see it confusing folks.

src/main/java/minecrafttransportsimulator/rendering/components/AnimationVariableEasing.java
DonBruce64: Couldn't they all go in the switch though?  None are more than 5 lines, so you could easily return some values directly as a 1-liner, and then others with a `{}` return area.

DonBruce64: But, I can see it getting a bit clunky.    Will redact this comment.

src/main/java/minecrafttransportsimulator/rendering/components/AnimationVariableEasing.java
```diff
@@ -0,0 +1,239 @@
...
+			case EASEINOUTBACK: return easeInOutBack(time);
+			case EASEINELASTIC: return easeInElastic(time);
+			case EASEOUTELASTIC: return easeOutElastic(time);
+			case EASEINOUTELASTIC: return easeInOutElastic(time);
+			case EASEINBOUNCE: return easeInBounce(time);
+			case EASEOUTBOUNCE: return easeOutBounce(time);
+			case EASEINOUTBOUNCE: return easeInOutBounce(time);
+			
+			//Easing type is invalid. Log error
+			default: {
```
DonBruce64, line 58: You won't ever get a default easing type.  If the enum isn't listed as a case, then that's an issue with the code missing an enum.  If all easing enums are listed, then there's literally no way to get an enum that isn't one of them as it literally does not exist.  If someone puts a wrong enum name in the JSON, then it'll flag that JSON for an error and the JSON won't load.

Also, that log message, even if it did occur, would log every time the animation was called.  So you'd end up with a line in your logger every tick.  Bad Time.

DonBruce64 (changes requested):

src/main/java/minecrafttransportsimulator/rendering/components/DurationDelayClock.java
RishiMenon2004: aight, i'll default it to linear

src/main/java/minecrafttransportsimulator/rendering/components/DurationDelayClock.java
RishiMenon2004: oh right-

src/main/java/minecrafttransportsimulator/rendering/components/AnimationVariableEasing.java
RishiMenon2004: oh ._. tried to do a good, instead did a dumb 😎

src/main/java/minecrafttransportsimulator/rendering/components/DurationDelayClock.java
RishiMenon2004: wait what

DonBruce64: I'm saying that time is equal to `timeMoved/(double)(animation.duration*50);`  But, before you call getEasingValue, you calculate `movementFactor = timeMoved/(double)(animation.duration*50);`  So just feed in movementFactor as the variable to this function so you don't calculate twice is what I'm saying.

RishiMenon2004: so i moved `movementFactor` to the function as `time`, so i could do ``movementFactor = getEasingValue(animation, timeMoved, false);``
instead of doing ``movementFactor = timeMoved/(double)(animation.duration*50);``
then again ``movementFactor = getEasingFactor(animation, timeMoved, false);``

src/main/java/minecrafttransportsimulator/rendering/components/AnimationVariableEasing.java
RishiMenon2004: so i can't remove the default?
image

PR #854, opened 2021-06-21
weightTransfer

BlazeWolf1122:
slightly more advanced version of overSteer that works independent of overSteer.
i'm thinking about making this the new overSteer code, but i'm on the fence about it.
If you know it's a significant  optimization to turn overSteer code into the weightTransfer, i'll do that


BlazeWolf1122:
image

PR #935, opened 2021-09-04
[PR] Definable Engine Shutoff Winddown Rate (or whatever the hell you call it) and Definable-By-Engine Hours Factor (which cannot go below 0.25)

boot2big:
By default slowdown-after-turning-off-ignition-or-stalling will still be 10, but can be changed to something like 50 to make engines shut off 5x as fast, or 5 to make them shut off 0.5x as fast.

Engine hours thing will be a step above config, but defined by engine and cannot go below 0.25

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,5 @@
+# MinecraftTransportSimulator-but-CC-fucked-it
+A Minecraft mod that adds planes and automobiles with realistic physics!
+
+
+Holy hell how do I use github aaaargh what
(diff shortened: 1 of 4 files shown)


boot2big:
Ignore first and last commits. Pain.

DonBruce64:
Did not ignore first and last commits.  Must see others pain to make my pain go away.  Smol question: why does the engine hours rate LC check if it's less than 0.25?  I would think it should check it if's zero like the other config?  Or is that to set a lower bounds to prevent OP Engine Pack 2.0?

boot2big:
Bingo, it's to prevent OP engines. Because in my testing, 0.25 was already fairly powerful, though it'd probably be best for electrics.

DonBruce64:
Looks good to me. In it goes!

PR #946, opened 2021-09-19
[PR] New engine hours behaviour + lugging

boot2big:
Engines don't sputter as quickly, doesn't lose power as fast. Lugging is less luggy unless sputtery.


src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
```diff
@@ -412,8 +413,11 @@ public boolean update(){
 						if(wheelFriction > 0){
 							double desiredRPM = lowestWheelVelocity*1200F*currentGearRatio*vehicleOn.definition.motorized.axleRatio;
 							rpm += (desiredRPM - rpm)/definition.engine.revResistance;
-							if(rpm < definition.engine.stallRPM && state.running){
+							if(rpm < definition.engine.idleRPM && state.running && backfireCooldown <= 0){//Checks if we're backfiring and sets lugging rpm to stall rpm, otherwise sets lug rpm to idle
+								rpm = definition.engine.idleRPM;
+							}else if(rpm < definition.engine.stallRPM && state.running){
 								rpm = definition.engine.stallRPM;
+								backfireCooldown -= 0.25;
```
DonBruce64, line 420: This should really be an int, then you check if it's greater than 0, and if so, do --backfireCooldown to make it go down by one.  Doing math on floating-points leads to fuzzes where occasionally the cooldown will get to 0.00000001, so you don't get consistent behavior.

src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
```diff
@@ -296,8 +297,8 @@ public boolean update(){
 						}
 						
 						//If the engine has high hours, give a chance for a backfire.
-						if(hours > 100 && !world.isClient()){
-							if(Math.random() < hours/1000*(definition.engine.maxSafeRPM/(rpm+definition.engine.maxSafeRPM/2))){
+						if(hours > 250 && !world.isClient()){
+							if(Math.random() < (hours/2)/(10250-hours)*(definition.engine.maxSafeRPM/(rpm+definition.engine.maxSafeRPM/1.5))){
```
DonBruce64, line 301: Given this math, if I get my engine above 10250 hours I won't ever back-fire, because I'll be dividing by a negative number so the right-hand side of the equation will be negative, and the random left-hand side will never be less than a negative number.  Food for thought.  Will hours ever get that high?  Prolly not.

src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
DonBruce64, line 420: This should really be an int, then you check if it's greater than 0, and if so, do --backfireCooldown to make it go down by one. Doing math on floating-points leads to fuzzes where occasionally the cooldown will get to 0.00000001, so you don't get consistent behavior.

DonBruce64 (changes requested):

src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
boot2big: Fixed by instead making it subtract from 10000, and just adding 250 to it. Also, the sputtering would probably kill it long before 10000 hours, maybe even before it gets halfway.

src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
boot2big: Fixed by... making it an int with the same behavior (only now it's 4, and it subtracts 1 every tick)

PR #1016, opened 2021-12-15
Add translations to physics modifiers

BlazeWolf1122:
oh shit i didn't think it was that easy
you weren't kidding
anyways, i suspect that i need to give the same treatment to parts as well or else physics modifier translations can't work on parts


BlazeWolf1122:
scratch that it was a painful process

DonBruce64:
I don't like the stat this is in, but I need it pulled in to do rotation overhaul work as it'll change this code.  Better to pull in now and do the work myself rather than let it linger.

PR #1035, opened 2021-12-31
Added speed_scaled

DrprofLuigi:
speed_scaled outputs the scaled speed of the vehicle. That way sounds that should only be heard at high speed are heard more at a higher SF setting. Also eliminates the need for multiple UNU speedometers :sans:


DonBruce64:
Code looks good, but don't you have the wiki description for the variable backwards?  speed_scaled isn't multiplied by the speed factor, least not what I can see?

DrprofLuigi:
Sorry, it divides by the speed factor. I got it switched in my explanation because it has the effect of multiplying to make up the difference when the SF is less than one.  The code is correct, but my wiki entry was incorrect (which I'll fix).

Issue #1085, opened 2022-02-07
[ENHANCEMENT] *Some* form of speedfactor workaround


boot2big:
@DrprofLuigi and @BlazeWolf1122 on this as well as pings in #garage so that we can agree on WTF this ticket actually means

I'm going to assume for now that it will refer to pack-defined speedfactors for planes and cars separately, but clearly defined discussion here would help.

Ultimately if the above isn't possible we can make do with an animation variable for speedfactor, but lets see if it _is_ possible first before going the variableModifiers route

boot2big:
As this topic doesn't seem to have been picked up much lately, I've decided to take some personal action regarding it...

Secondary solution that I had proposed
![image](https://user-images.githubusercontent.com/53581325/154510088-f9243923-6932-48f4-9e5a-18f604891bd0.png)
![image](https://user-images.githubusercontent.com/53581325/154510062-c2e0d31b-3c51-4d68-833e-b51c72e1c3e9.png)
![image](https://user-images.githubusercontent.com/53581325/154510346-10df55d3-95a3-4c95-bd77-63abc712b520.png)

boot2big:
I'm going to keep my eye on this ticket for a few more days, and if nothing else is proposed and/or alternative implementation isn't started by then, I'll close this ticket and assume that VMs+speed_factor is the de-facto method

DonBruce64:
Oooor, you wait for me to get out of V21 coding hell and let me address this properly.

DonBruce64:
Added in V21.0.0.

PR #1099, opened 2022-02-17
[Minor PR] speed_factor Variable

boot2big:
If nothing else is decided in #1085, this animation variable would allow for PAs to account for speedFactor accordingly, in all aspects (Top speed/acceleration, braking, handling, ETC)


DonBruce64:
I see no reason why this can't exist in the current code in the interim.  Approved.  Just make sure to update the wiki page.

PR #1160, opened 2022-04-11
Feature request: Add a pipeline to build (validate) the project on every commit

recolic:
Benefits: 
1. If a new commit causes compilation error, we can immediately know it and fix it. 
2. If someone complains that he can not build this project, we can say "it's your problem, not ours". 
3. It automatically publishes build result for every commit. So the user can directly download the latest nightly build, and immediately try it.

### DIFF
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -1,0 +1,36 @@
+name: Build
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


DonBruce64:
I like this idea.  Lemme get this next release out though before I pull this in, as there's a lot of kerfuffle about the various bugs, and I want em closed out before I poke the building system with a stick.

DonBruce64:
Looking into this here I noticed that there's mention of Ubuntu.  Does that mean this system requires a server somewhere to work?  I've not ever messed with auto-build systems.

recolic:
> Looking into this here I noticed that there's mention of Ubuntu. Does that mean this system requires a server somewhere to work? I've not ever messed with auto-build systems.

No. We says `runs-on: ubuntu-latest` in the yml, and github will run our pipeline on ubuntu environment. 
Github will take care of the server management, and we need to do nothing about that.

Issue #1278, opened 2022-08-06
[ENHANCEMENT] Birdshot/Spray/Pellet ammo that shoots multiple of itself per single shot


boot2big:
Random trajectories that differ from eachother so that a gun with a proper bulletSpreadFactor set will have each "bullet" (pellet) fire at a different angle, and amount of pellets can be set (with 0 being LCd to one, for a single bullet, or possibly the variable being an additive thing meaning 0 adds 0 bullets, 4 gives a total of five shots per round, etc)

DonBruce64:
BSF is for the firing from the gun, not for actual spread.  What you're looking for is some system to fire off multiple bullets per shot.  Granted, a shotgun will probably use the BSF for each of those pellets, but there has to be a "count" of some sort to tell the gun to spawn all those bullets in the first place.  So for sure add that variable and LC it to 1, but no need to poke BSF.

ktokto313:
I got a working beta for this on the-open-road, Laura and I both performed individual test on the pack and have seen no bug or error, I am currently putting a pull request for that.

DonBruce64:
Implemented in V22.0.0

PR #1282, opened 2022-08-07
Shotgun bullet implement according to issue #1278

ktokto313:
As the title says and a little patch for implement bullet "hardness" for bullet breaking glass only which can be removed later on

### DIFF
--- a/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
+++ b/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
@@ -290,12 +290,16 @@ public class PartGun extends APart {
                             }
 
                             if (loadedBullet.definition.bullet.pellets > 1) {
+                                //Spawn all pellet into the world
                                 for (int i=0;i<loadedBullet.definition.bullet.pellets;i++) {
+                                    //Give pellet random spread based on the bore sight
+                                    Point3D randomBulletVelocity = bulletVelocity;
                                     if (loadedBullet.definition.bullet.pelletSpreadFactor > 0) {
-                                        firingSpreadRotation.angles.set((Math.random() - 0.5F) * loadedBullet.definition.bullet.pelletSpreadFactor, (Math.random() - 0.5F) * loadedBullet.definition.bullet.pelletSpreadFactor, 0D);
-                                        bulletVelocity.rotate(firingSpreadRotation);
+
+                                        firingSpreadRotation.angles.set((Math.random() - 0.5F) * loadedBullet.definition.bullet.pelletSpreadFactor, (Math.random() - 0.5F) * loadedBullet.definition.bullet.pelletSpreadFactor, (Math.random() - 0.5F) * loadedBullet.definition.bullet.pelletSpreadFactor);
+                                        randomBulletVelocity.rotate(firingSpreadRotation);
                                     }
-                                    newBullet = new EntityBullet(bulletPosition, bulletVelocity, bulletOrientation, this);
+                                    newBullet = new EntityBullet(bulletPosition, randomBulletVelocity, bulletOrientation, this);
                                     world.addEntity(newBullet);
                                 }
                             } else
@@ -712,7 +716,7 @@ public class PartGun extends APart {
             bulletVelocity.add(motion);
         }
 
-        //Set position.
+        //Set bullet position.
         bulletPosition.set(muzzle.pos);
         if (muzzle.center != null) {
             pitchMuzzleRotation.setToZero().rotateX(internalOrientation.angles.x);
@@ -723,7 +727,7 @@ public class PartGun extends APart {
         }
         bulletPosition.rotate(zeroReferenceOrientation).add(position);
 
-        //Set orientation.
+        //Set bullet orientation.
         bulletOrientation.set(zeroReferenceOrientation).multiply(internalOrientation);
         if (muzzle.rot != null && !definition.gun.disableMuzzleOrientation) {
             bulletOrientation.multiply(muzzle.rot);
--- a/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
+++ b/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
@@ -289,6 +289,16 @@ public class PartGun extends APart {
                                 newBullet = new EntityBullet(bulletPosition, bulletVelocity, bulletOrientation, this);
                             }
 
+                            if (loadedBullet.definition.bullet.pellets > 0) {
+                                for (int i=0;i<loadedBullet.definition.bullet.pellets;i++) {
+                                    if (definition.gun.bulletSpreadFactor > 0) {
+                                        firingSpreadRotation.angles.set((Math.random() - 0.5F) * definition.gun.bulletSpreadFactor, (Math.random() - 0.5F) * definition.gun.bulletSpreadFactor, 0D);
+                                        bulletVelocity.rotate(firingSpreadRotation);
+                                    }
+                                    newBullet = new EntityBullet(bulletPosition, bulletVelocity, bulletOrientation, this);
+                                    world.addEntity(newBullet);
+                                }
+                            } else
                             world.addEntity(newBullet);
 
                             //Decrement bullets, but check to make sure we still have some.
--- a/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
+++ b/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
@@ -291,7 +291,7 @@ public class PartGun extends APart {
 
                             if (loadedBullet.definition.bullet.pellets > 1) {
                                 for (int i=0;i<loadedBullet.definition.bullet.pellets;i++) {
-                                    if (definition.gun.bulletSpreadFactor > 0) {
+                                    if (loadedBullet.definition.bullet.pelletSpreadFactor > 0) {
                                         firingSpreadRotation.angles.set((Math.random() - 0.5F) * loadedBullet.definition.bullet.pelletSpreadFactor, (Math.random() - 0.5F) * loadedBullet.definition.bullet.pelletSpreadFactor, 0D);
                                         bulletVelocity.rotate(firingSpreadRotation);
                                     }
--- a/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
+++ b/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
@@ -683,7 +683,7 @@ public class PartGun extends APart {
         if (definition.gun.muzzleVelocity != 0) {
             bulletVelocity.set(0, 0, definition.gun.muzzleVelocity / 20D / 10D);
             //Randomize the spread for normal bullet and pellets
-            if (definition.gun.bulletSpreadFactor > 0) {
+            if (definition.gun.bulletSpreadFactor > 0 || loadedBullet.definition.bullet.pelletSpreadFactor > 0) {
                 firingSpreadRotation.angles.set((Math.random() - 0.5F) * (definition.gun.bulletSpreadFactor + loadedBullet.definition.bullet.pelletSpreadFactor), (Math.random() - 0.5F) * (definition.gun.bulletSpreadFactor + loadedBullet.definition.bullet.pelletSpreadFactor), 0D);
                 bulletVelocity.rotate(firingSpreadRotation);
             }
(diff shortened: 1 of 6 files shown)


ktokto313:
Why is the log, still printing welcome to beta 7 lol, can someone fix it for me?

src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -289,6 +289,33 @@ public void update() {
...
                             }
 
+                            if (loadedBullet.definition.bullet.pellets > 1) {
+                                //Spawn all pellet into the world
+                                for (int i=0;i<loadedBullet.definition.bullet.pellets;i++) {
+                                    //Give pellet random spread based on the bore sight
+                                    Point3D randomBulletVelocity = bulletVelocity;
+                                    if (loadedBullet.definition.bullet.pelletSpreadFactor > 0) {
+                                        firingSpreadRotation.angles.set((Math.random() - 0.5F) * loadedBullet.definition.bullet.pelletSpreadFactor, (Math.random() - 0.5F) * loadedBullet.definition.bullet.pelletSpreadFactor, (Math.random() - 0.5F) * loadedBullet.definition.bullet.pelletSpreadFactor);
+                                        randomBulletVelocity.rotate(firingSpreadRotation);
```
DonBruce64, line 299: This won't work here.  You need to apply the random rotation before you get the actual velocity.  The rotation applied is relative to the orientation of the gun, and gets applied before the matrix rotations that orient the velocity to the gun.  This should be in "setBulletSpawn" with the bullet spread, not here.

src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -289,6 +289,33 @@ public void update() {
                                 newBullet = new EntityBullet(bulletPosition, bulletVelocity, bulletOrientation, this);
                             }
 
+                            if (loadedBullet.definition.bullet.pellets > 1) {
+                                //Spawn all pellet into the world
+                                for (int i=0;i<loadedBullet.definition.bullet.pellets;i++) {
+                                    //Give pellet random spread based on the bore sight
+                                    Point3D randomBulletVelocity = bulletVelocity;
```
DonBruce64, line 296: Don't do this.  This doesn't copy the variable.  This just takes it and changes its name in the middle of the code.  That's just confusion.  So really, you're editing bulletVelocity in all the calls below.

src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -289,6 +289,33 @@ public void update() {
...
+                                            //No entity found, just fire missile off in direction facing.
+                                            newBullet = new EntityBullet(bulletPosition, randomBulletVelocity, bulletOrientation, this);
+                                        }
+                                    } else {
+                                        //Normal bullet
+                                        newBullet = new EntityBullet(bulletPosition, randomBulletVelocity, bulletOrientation, this);
+                                    }
+                                    world.addEntity(newBullet);
+                                }
+                            } else
```
DonBruce64, line 318: You do an else here to do the normal spawn, but this means that you still create new bullets on lines 279 though 290, just you ignore those bullets you created and re-create them.  That's bad, as it wastes processing power.

src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
DonBruce64: FYI, changing the logic for this will make all other comments not matter much.  Other comments are for your information.

DonBruce64 (changes requested):

ktokto313:
@DonBruce64 Here is the fix, i tried to redo it as you say, there is still flaw in it imo, i hope i get further opinion of yours!

src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -270,26 +270,28 @@ public void update() {
...
+                                    } else {
+                                        //No entity found, just fire missile off in direction facing.
+                                        newBullet = new EntityBullet(bulletPosition, bulletVelocity, bulletOrientation, this);
+                                    }
                                 } else {
-                                    //No entity found, just fire missile off in direction facing.
                                     newBullet = new EntityBullet(bulletPosition, bulletVelocity, bulletOrientation, this);
                                 }
-                            } else {
-                                newBullet = new EntityBullet(bulletPosition, bulletVelocity, bulletOrientation, this);
```
DonBruce64, line 289: This bullet won't go the right way.  It never gets its spawn set.

src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -270,26 +270,28 @@ public void update() {
...
-                                } else if (engineTarget != null) {
-                                    newBullet = new EntityBullet(bulletPosition, bulletVelocity, bulletOrientation, this, engineTarget);
+                            for (int i=0; i < loadedBullet.definition.bullet.pellets; i++) {
+                                //Get the bullet's state.
+                                setBulletSpawn(bulletPosition, bulletVelocity, bulletOrientation, muzzle);
+
+                                //Add the bullet to the world.
+                                //If the bullet is a missile, give it a target.
+                                EntityBullet newBullet;
+                                if (loadedBullet.definition.bullet.turnRate > 0) {
```
DonBruce64, line 280: There's really no reason to check turn rate for pellets.  They will never be missiles.

src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -680,10 +682,16 @@ public void setBulletSpawn(Point3D bulletPosition, Point3D bulletVelocity, Rotat
...
             bulletVelocity.set(0, 0, definition.gun.muzzleVelocity / 20D / 10D);
+            //Randomize the spread for normal bullet
             if (definition.gun.bulletSpreadFactor > 0) {
                 firingSpreadRotation.angles.set((Math.random() - 0.5F) * definition.gun.bulletSpreadFactor, (Math.random() - 0.5F) * definition.gun.bulletSpreadFactor, 0D);
                 bulletVelocity.rotate(firingSpreadRotation);
             }
+            //Randomize the spread for shotgun shell pellets
+            if (loadedBullet.definition.bullet.pelletSpreadFactor > 0) {
+                firingSpreadRotation.angles.set((Math.random() - 0.5F) * loadedBullet.definition.bullet.pelletSpreadFactor, (Math.random() - 0.5F) * loadedBullet.definition.bullet.pelletSpreadFactor, 0D);
+                bulletVelocity.rotate(firingSpreadRotation);
```
DonBruce64, line 693: Due to the property of rotations, this won't work if the gun has a bullet spread factor.  I would say to add both the spread's X and Y components, and then rotate, but you may not want to use the gun's bullet spread with pellets.  Since that'll just double-up on the spread values.  Would be a @LauraDarkez question.

DonBruce64 (changes requested):

src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
boot2big: I would like to comment on that too though, afaik shorter barrels will make the pellets spread out more than longer barrels, so maybe BSF should have an effect alongside PSF

src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
ktokto313: why not, they can make sci-fi shotgun now :(

src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
ktokto313: Yes, thank you for reminding me that, i always work too fast and it turned out to be dumb and inefficient.

src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
ktokto313: Yes, I am not fully understood how vector 3d work, that's totally not my intention

DonBruce64: If the spread should be cumulative, then we can make it so.  But that's only if that is the case.  @Ilya-torshov gets a ping here too about this.

src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
LauraDarkez: i know that ive contemplated making a shotgun with turnrate pellets. and naomi had too

src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
LauraDarkez: okay now that ive seen this part of the git i know what to respond to. personally when i was trying to explain in #general, i tried to explain that the pelletSpreadFactor should ONLY be used when pellets > 0, and should NOT use bulletSpreadFactor. seems that was lost in translation somewhere

LauraDarkez: even though it could be accepted, one big issue is that it sends pellets flying in random directions but with a randomized overall vector, in reality i dont think this should be the case. it should randomly distribute the pellets but it should do so with a constant overall straight-forward direction. so sometimes pellets will cluster to the left but most of the time itll be a relative spread.

DonBruce64: By the effect of randomness, and the sheer amount of pellets, spread should be fairly uniform.  @ktokto313 it has been decided that the spread needs to be cumulative.  So you'll need to add both those random X and Z components and do a single rotation operation on them, rather than doing two, one after the other, as it will return bad values.

src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -680,8 +682,9 @@ public void setBulletSpawn(Point3D bulletPosition, Point3D bulletVelocity, Rotat
         //Set velocity.
         if (definition.gun.muzzleVelocity != 0) {
             bulletVelocity.set(0, 0, definition.gun.muzzleVelocity / 20D / 10D);
+            //Randomize the spread for normal bullet and pellets
             if (definition.gun.bulletSpreadFactor > 0) {
-                firingSpreadRotation.angles.set((Math.random() - 0.5F) * definition.gun.bulletSpreadFactor, (Math.random() - 0.5F) * definition.gun.bulletSpreadFactor, 0D);
+                firingSpreadRotation.angles.set((Math.random() - 0.5F) * (definition.gun.bulletSpreadFactor + loadedBullet.definition.bullet.pelletSpreadFactor), (Math.random() - 0.5F) * (definition.gun.bulletSpreadFactor + loadedBullet.definition.bullet.pelletSpreadFactor), 0D);
```
DonBruce64, line 687: Since these are factors, they should multiply, not add.  But watch out for one or the other being 0.

src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -680,8 +682,9 @@ public void setBulletSpawn(Point3D bulletPosition, Point3D bulletVelocity, Rotat
         //Set velocity.
         if (definition.gun.muzzleVelocity != 0) {
             bulletVelocity.set(0, 0, definition.gun.muzzleVelocity / 20D / 10D);
+            //Randomize the spread for normal bullet and pellets
             if (definition.gun.bulletSpreadFactor > 0) {
```
DonBruce64, line 686: This only lets the spread activate if the gun has a spread factor.  If the gun doesn't, but the bullet does, this code won't run, despite it needing to.

DonBruce64 (changes requested):

src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
ktokto313: It's basically (Math.random() - 0.5d) * bulletSpreadFactor + (Math.random() - 0.5d) * pelletSpreadFactor, i just reduce it

ktokto313:
@DonBruce64 the randomization code work alright for me, it's just the reduction of a multiplication, the condition to run the randomization code has been fixed

src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
DonBruce64: Apparently, I'm the one who can't do math today...

DonBruce64 (approved):

Issue #1307, opened 2022-08-31
Remove Y (up down)axis jet engine force


Naxgeneral:
since it mess up a lot so it should be gone

DonBruce64:
Fixed in V21.1.0.

PR #1311, opened 2022-09-01
Remove X-axis torque and prevent user pressing both A and D simutaniously

ktokto313:
Remove X-axis torque according to issue #1307


ktokto313:
I need to poke the github rebase thingy, so i gonna kill this pr

PR #1330, opened 2022-09-13
Added 'config_simplethrottle'

DrprofLuigi:
The goal here is to add the 'config_simplethrottle' variable to AEntityD_Definable, but I think I screwed up the github step because the PR seems to be doing something to the force shift VM.

I spent half an hour trying to figure out how to remove those commits, but I don't know what the heck I'm doing when it comes to github


DrprofLuigi:
Once this is pulled in, I'll add config_simplethrottle to the wiki.

DonBruce64:
Not sure if it was intended, but you removed forceShift entierly with this PR.  Pretty sure that breaks your transmissions and you don't want that.

DrprofLuigi:
> Not sure if it was intended, but you removed forceShift entierly with this PR. Pretty sure that breaks your transmissions and you don't want that.

That was not intended, it's my ineptitude with github. I couldn't remove that commit, I don't even know how it got there.

PR #1333, opened 2022-09-14
Added config_simplethrottle

DrprofLuigi:
This is a simple value that returns config_simplethrottle as a 1 or 0 depending on the config value.


DonBruce64:
Looks good.  Just make sure to update the wiki with the new variable.

DrprofLuigi:
> Looks good. Just make sure to update the wiki with the new variable.

Done

PR #1337, opened 2022-09-21
Tweaked engine variables (with some additional functionality)

DrprofLuigi:
Fixed `engine_piston` variable to be more intuitive and consistent.
Added `_cam` suffix to specify using cam rotation (two rotations of the engine).
Added the ability to offset the engine rotation input for `engine_sin` and `engine_cos` variables for more functionality.
Tweaked the README file to include a git step.


DrprofLuigi:
I'll update the wiki when this is pulled in.

PR #1344, opened 2022-09-30
Added the ability to offset the piston variable

DrprofLuigi:
This is similar to how engine trig variables can be offset. I'll update the wiki when the PR is approved.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,20 @@
+# common
+run
+build
+.gradle
+
+# eclipse
+eclipse
+bin
+*.launch
+.metadata
+.classpath
+.project
+.settings
+
+# idea
+out
+*.ipr
+*.iws
+*.iml
+.idea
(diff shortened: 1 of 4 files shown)


src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
```diff
@@ -835,15 +835,25 @@ public double getRawVariableValue(String variable, float partialTicks) {
...
+        	//Extract the values we need
             String pistonVariable = variable.substring("engine_piston_".length());
-            int pistonNumber = Integer.parseInt(pistonVariable.substring(0, pistonVariable.indexOf("_")));
-            int totalPistons = Integer.parseInt(pistonVariable.substring(pistonVariable.indexOf("_") + 1, pistonVariable.indexOf("_", pistonVariable.indexOf("_") + 1)));
+            String[] parsedVariable = pistonVariable.split("_");
+            int pistonNumber = Integer.parseInt(parsedVariable[0]);
+            int totalPistons = Integer.parseInt(parsedVariable[1]);
+        	int offset = 0;
+        	//Try to assign an offset if it is present
+            try {
```
DonBruce64, line 852: You don't want to do a try-catch here, that takes a lot of CPU time.  Instead, check if the length of parsedVariable `parsedVariable.length` is 3.

DonBruce64 (changes requested):

Issue #1368, opened 2022-10-13
When parts with default subParts are replaced, default subParts are missing.


DrprofLuigi:
If I pull off and re-place an UNU MG that has a default and permanent invisiseat, the invisiseat is missing.

This is neither caused by nor fixed with PR #1369

DonBruce64:
Fixed in 22.3.0.

PR #1369, opened 2022-10-13
Added lockingVariables

DrprofLuigi:
lockingVariables act just like interactableVariables. However, if the list is true, the part is 'locked' and acts like a permanent part.

Note: this PR neither causes or fixes issue #1368


DrprofLuigi:
This PR is too far behind, I'll have to redo it.

Issue #1356, opened 2022-10-07
[BUG] engine_piston_'s new functionality requires more clarification on the wiki, or is broken.


boot2big:
_"The system is down. I dunno what you did, moron, but you sure screwed everything up good."_
![image](https://user-images.githubusercontent.com/53581325/194670236-6e14522f-9c4c-40e6-8bf1-3156ac2c3dce.png)
![image](https://user-images.githubusercontent.com/53581325/194670265-16545b00-d0b0-44b2-a0dd-53578b442079.png)
[crash-2022-10-07_18.14.37-client.txt](https://github.com/DonBruce64/MinecraftTransportSimulator/files/9737713/crash-2022-10-07_18.14.37-client.txt)

Either the system for registering that animation variable is borked, or whoever made the wiki entry didn't test that example before putting it there. Whichever one, that møøse needs to get sacked!

DonBruce64:
It would appear there needs to be some sort of suffix on that variable.  Indeed, variables with a _X suffix designate part-specific variables, of which the engine_pistion variable is not.  So there is a wiki entry there.

DrprofLuigi:
Are you using that variable on a vehicle, or a part?

Just like every other variable listed on the wiki, you'll need to suffix the part number if you are referencing a part with a part variable. 

With `engine_gearshift`, if you are using it on a vehicle to reference a part, you need to make it `engine_gearshift_1` So if you are using that on a vehicle, you need it to be `engine_piston_2_3_1` to return `engine_piston_2_3` from engine number 1.

The wiki entry is consistent with how all other variables are listed.

DonBruce64:
I'm going to refute the claim of consistent naming.  As per the Animation Hierarchy page.

`The first thing MTS checks is if the animation has a number on the end of it. If so, MTS will try to get the part of that number as defined in the order of the [part slots](https://github.com/DonBruce64/MinecraftTransportSimulator/wiki/Pack-Making-JSON-Part-Slots). So, if you have a variable engine_rotation_2, MTS will attempt to get the second engine and return the value of engine_rotation for that engine. The same goes for things like gun_ammo_1 to get the ammo count of the first gun and propeller_pitch_percent_3 to get the propeller pitch of the third propeller, and part_present_22 to check if the 22nd part is present. If MTS can't find the part with the specified index, it will assume that it simply hasn't been placed yet, and will return a value of 0 for the animation.`

[...]

PR #1390, opened 2022-10-22
Added the ability to control part permanent-ness

DrprofLuigi:
Also fixed normalized wheel rotation, and modified the engine_piston variable to be compliant as per #1356.


mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityF_Multipart.java
```diff
@@ -380,6 +380,42 @@ public boolean areVariablesBlocking(JSONPartDefinition partDef, IWrapperPlayer p
         return false;
     }
 
+    /**
+     * Returns true if any lockingVariables are locking the passed-in part slot.
+     */
+    public boolean areVariablesLocking(JSONPartDefinition partDef, IWrapperPlayer player, boolean defaultValue) {
+        if (partDef.lockingVariables != null) {
+            for (List<String> variableList : partDef.lockingVariables) {
```
DonBruce64, line 388: Looking at this, I only see this variable being different vs the linkedVariables, plus the default.  In this case, the better option is to take the areVaraiblesBlocking function and add the variableList as a paramter so they run off the same code.  Helps if something borks.  As to your defaultValue, that doesn't need to be in here, as you should first check if lockingVariables == null before calling this method, and if it is, just use the default.  I can help clarify, but I say this given what I had to do with the AnimationSwitchbox code.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
```diff
@@ -160,6 +161,9 @@ public void update() {
         position.set(entityOn.position);
         orientation.set(entityOn.orientation);
         localOffset.set(placementDefinition.pos);
+        
+        //Update permanent-ness
+        isPermanent = areVariablesLocking(placementDefinition, InterfaceManager.clientInterface.getClientPlayer(),placementDefinition.isPermanent);
```
DonBruce64, line 166: `InterfaceManager.clientInterface.getClientPlayer(`  BAD.  this can't run on the server.  You don't need the player anyways in the code function either.  Though I see you left that since the player was in the linkedVariables code.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
```diff
@@ -909,7 +912,8 @@ public double getRawVariableValue(String variable, float partialTicks) {
             }
             
             //Map the shaft rotation to a value between 0 and 359.99...
-            double shaftRotation = (offset + getEngineRotation(partialTicks)) - ((360D * camMultiplier) * Math.floor((offset + getEngineRotation(partialTicks)) / (360D * camMultiplier)));
+            //double shaftRotation = (offset + getEngineRotation(partialTicks)) - ((360D * camMultiplier) * Math.floor((offset + getEngineRotation(partialTicks)) / (360D * camMultiplier)));
```
DonBruce64, line 915: Code not do code here.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGroundDevice.java
```diff
@@ -233,7 +233,7 @@ public double getRawVariableValue(String variable, float partialTicks) {
             case ("ground_rotation"):
                 return vehicleOn != null ? vehicleOn.speedFactor * (partialTicks != 0 ? prevAngularPosition + (angularPosition - prevAngularPosition) * partialTicks : angularPosition) * 360D : 0;
             case ("ground_rotation_normalized"):
-                return vehicleOn != null ? (vehicleOn.speedFactor * (partialTicks != 0 ? prevAngularPosition + (angularPosition - prevAngularPosition) * partialTicks : angularPosition) * 360D) % 360D : 0;
+            	return vehicleOn != null ? Math.floorMod(Math.round(vehicleOn.speedFactor * (prevAngularPosition + (angularPosition - prevAngularPosition) * partialTicks) * 3600), 3600) / 10D : 0;
```
DonBruce64, line 236: You shouldn't have to round anything here....?  If it works, it works, but I think there's bad maths here somewhere.  Normalization never requires rounding.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGroundDevice.java
DrprofLuigi: Math.floorMod told me it needed integers. Plus making it tenths makes it easier for clamping which is one of the reasons this was added. Though it may be better to have a piston-like version in the future.

Math.floorMod is needed because it allows for the modulo to work as expected with negative values.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
DrprofLuigi: I removed that, I kept it in accidentally.

mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityF_Multipart.java
DrprofLuigi: I combined them into `isVariableListTrue`, see my notes in the commit for details. It should be satisfactory, let me know if there are issues.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
DrprofLuigi: I removed it from the combined function `isVariableListTrue` since `player` wasn't used in the function anyway.

mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityF_Multipart.java
DonBruce64: Code looks fine for that.  Though the auto-build-checker script is saying that you missed a function call?  Or is it just obsolete?
`/java/minecrafttransportsimulator/entities/instances/APart.java:424: error: cannot find symbol
        return !entityOn.areVariablesBlocking(placementDefinition, InterfaceManager.clientInterface.getClientPlayer());`

DrprofLuigi: I guess I missed that one, it's fixed now.

PR #1393, opened 2022-10-28
Added Screwdriver Item

DrprofLuigi:
A Screwdriver is simply a mutually exclusive wrench that can only remove parts tagged with `mustBeRemovedByScrewdriver` in their generic section. Wrenches will ignore these parts.

Perfect for things like hubcaps where you may want to remove the wheel as an assembly when paired with `allowParentRemoval`.

This _should_ add a screwdriver item to the MTS core mod. It's even red!

Currently wrenches ONLY remove parts. They have no other functionality like a wrench does, but they could in the future.

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
@@ -255,25 +255,27 @@ public abstract class APart extends AEntityF_Multipart<JSONPart> {
             return;
         }
 
-    	boolean isHoldingWrench = world.isClient() ? InterfaceManager.clientInterface.getClientPlayer().isHoldingItemType(ItemComponentType.WRENCH) : false;
-    	boolean isHoldingScrewdriver = world.isClient() ? InterfaceManager.clientInterface.getClientPlayer().isHoldingItemType(ItemComponentType.SCREWDRIVER) : false;
-
-        //If we are holding a screwdriver or wrench, run these checks to remove hitboxes if needed.
-        if (world.isClient() && (isHoldingWrench || isHoldingScrewdriver)) {
-            //If we are holding a wrench and the part requires a screwdriver, remove interaction boxes so they don't get in the way and vice versa.
-            if ((isHoldingWrench && definition.generic.mustBeRemovedByScrewdriver) || (isHoldingScrewdriver && !definition.generic.mustBeRemovedByScrewdriver)) {
-                allInteractionBoxes.removeAll(interactionBoxes);
-                return;
-            }
-            //If we are holding a wrench or screwdriver, and the part has children, don't add the interaction boxes.  We can't wrench those parts.
-            //The only exceptions are parts that have permanent-default parts on them. or if they specifically don't block subpart removal.  These can be removed.
-            //Again, this only applies on clients for that client player.
-        	for (APart childPart : parts) {
-                if (!childPart.isPermanent && !childPart.placementDefinition.allowParentRemoval) {
-                    allInteractionBoxes.removeAll(interactionBoxes);
-                    return;
-                }
-            }
+        //If we are holding a screwdriver or wrench, run these checks to remove hitboxes if needed. This can only be done on the client.
+        if (world.isClient()) {
+	    	boolean isHoldingWrench = InterfaceManager.clientInterface.getClientPlayer().isHoldingItemType(ItemComponentType.WRENCH);
+	    	boolean isHoldingScrewdriver = InterfaceManager.clientInterface.getClientPlayer().isHoldingItemType(ItemComponentType.SCREWDRIVER);
+	
+	        if (isHoldingWrench || isHoldingScrewdriver) {
+	            //If we are holding a wrench and the part requires a screwdriver, remove interaction boxes so they don't get in the way and vice versa.
+	            if ((isHoldingWrench && definition.generic.mustBeRemovedByScrewdriver) || (isHoldingScrewdriver && !definition.generic.mustBeRemovedByScrewdriver)) {
+	                allInteractionBoxes.removeAll(interactionBoxes);
+	                return;
+	            }
+	            //If we are holding a wrench or screwdriver, and the part has children, don't add the interaction boxes.  We can't wrench those parts.
+	            //The only exceptions are parts that have permanent-default parts on them. or if they specifically don't block subpart removal.  These can be removed.
+	            //Again, this only applies on clients for that client player.
+	        	for (APart childPart : parts) {
+	                if (!childPart.isPermanent && !childPart.placementDefinition.allowParentRemoval) {
+	                    allInteractionBoxes.removeAll(interactionBoxes);
+	                    return;
+	                }
+	            }
+	        }
         }
     }
 
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
@@ -255,32 +255,16 @@ public abstract class APart extends AEntityF_Multipart<JSONPart> {
             return;
         }
 
-        //If we are holding a wrench, and the part has children, don't add the interaction boxes.  We can't wrench those parts.
-        //The only exceptions are parts that have permanent-default parts on them. or if they specifically don't block subpart removal.  These can be wrenched.
-        //Again, this only applies on clients for that client player.
-        if (world.isClient() && InterfaceManager.clientInterface.getClientPlayer().isHoldingItemType(ItemComponentType.WRENCH)) {
-        	//If we are holding a screwdriver, remove the interaction boxes so they don't get in the way.
-            if (definition.generic.mustBeRemovedByScrewdriver) {
+        //If we are holding a screwdriver or wrench, run these checks to remove hitboxes if needed.
+        if (world.isClient() && (InterfaceManager.clientInterface.getClientPlayer().isHoldingItemType(ItemComponentType.WRENCH) || InterfaceManager.clientInterface.getClientPlayer().isHoldingItemType(ItemComponentType.SCREWDRIVER))) {
+            //If we are holding a wrench and the part requires a screwdriver, remove interaction boxes so they don't get in the way and vice versa.
+            if ((InterfaceManager.clientInterface.getClientPlayer().isHoldingItemType(ItemComponentType.WRENCH) && definition.generic.mustBeRemovedByScrewdriver) || (InterfaceManager.clientInterface.getClientPlayer().isHoldingItemType(ItemComponentType.SCREWDRIVER) && !definition.generic.mustBeRemovedByScrewdriver)) {
                 allInteractionBoxes.removeAll(interactionBoxes);
                 return;
             }
-            for (APart childPart : parts) {
-                if (!childPart.isPermanent && !childPart.placementDefinition.allowParentRemoval) {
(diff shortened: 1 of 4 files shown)


mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONPart.java
```diff
@@ -93,6 +93,9 @@ public static class JSONPartGeneric {
         @JSONDescription("If true, this part will be able to be removed by hand and without a wrench.  This also bypasses owner requirements (but not vehicle locking).  Useful for small parts like luggage that anyone should be able to remove at any time.")
         public boolean canBeRemovedByHand;
 
+        @JSONDescription("If true, this part will be able to be removed by hand and without a wrench.  This also bypasses owner requirements (but not vehicle locking).  Useful for small parts like luggage that anyone should be able to remove at any time.")
```
DonBruce64, line 96: Description is incorrect.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
```diff
@@ -259,13 +259,35 @@ protected void updateEncompassingBox() {
         //The only exceptions are parts that have permanent-default parts on them. or if they specifically don't block subpart removal.  These can be wrenched.
         //Again, this only applies on clients for that client player.
         if (world.isClient() && InterfaceManager.clientInterface.getClientPlayer().isHoldingItemType(ItemComponentType.WRENCH)) {
+        	//If we are holding a screwdriver, remove the interaction boxes so they don't get in the way.
```
DonBruce64, line 262: This comment doesn't match the logic.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
```diff
@@ -259,13 +259,35 @@ protected void updateEncompassingBox() {
...
                     return;
                 }
             }
         }
+
+        //If we are holding a screwdriver, and the part has children, don't add the interaction boxes.  We can't screwdriver those parts.
+        //The only exceptions are parts that have permanent-default parts on them. or if they specifically don't block subpart removal.  These can be screwdrivered.
+        //Again, this only applies on clients for that client player.
+        if (world.isClient() && InterfaceManager.clientInterface.getClientPlayer().isHoldingItemType(ItemComponentType.SCREWDRIVER)) {
+            //If we are holding a wrench, remove the interaction boxes so they don't get in the way.
```
DonBruce64, line 279: But you aren't holding a wrench according to this `if` statement.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
```diff
@@ -259,13 +259,35 @@ protected void updateEncompassingBox() {
...
+            if (!definition.generic.mustBeRemovedByScrewdriver) {
+                allInteractionBoxes.removeAll(interactionBoxes);
+                return;
+            }
+        	for (APart childPart : parts) {
```
DonBruce64, line 284: I'm not sure if you need this check.  This check was designed to prevent removal of parts with parts on them.  Pretty sure screwdriverable parts won't have parts on them themselves?

mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
```diff
@@ -259,13 +259,35 @@ protected void updateEncompassingBox() {
...
                     allInteractionBoxes.removeAll(interactionBoxes);
                     return;
                 }
             }
         }
+
+        //If we are holding a screwdriver, and the part has children, don't add the interaction boxes.  We can't screwdriver those parts.
+        //The only exceptions are parts that have permanent-default parts on them. or if they specifically don't block subpart removal.  These can be screwdrivered.
+        //Again, this only applies on clients for that client player.
+        if (world.isClient() && InterfaceManager.clientInterface.getClientPlayer().isHoldingItemType(ItemComponentType.SCREWDRIVER)) {
```
DonBruce64, line 278: You could probably put this under the same if-block as the above.  First check if on client, then check if screwdriver or wrench (store as boolean), then check either boolean, and run common logic.  Cleaner than two individual code blocks with same code.

mccore/src/main/java/minecrafttransportsimulator/items/instances/ItemItem.java
```diff
@@ -54,72 +54,107 @@ public ItemItem(JSONItem definition) {
...
+	                if (!ownerState.equals(PlayerOwnerState.USER)) {
+	                    if (rightClick) {
+	                        //Right-clicking
+	                    } else {
+	                        //Left clicking removes parts, or X, if we were sneaking.
+	                        if(player.isSneaking()) {
+	                        }else {
+	                            if (entity instanceof APart) {
+	                                APart part = (APart) entity;
+	                                if (!part.isPermanent && part.isValid && part.definition.generic.mustBeRemovedByScrewdriver) {
```
DonBruce64, line 135: Given how this seems to be the only code difference between these blocks, what you should do is combine the case statements.  You can do this by writing:
```
case WRENCH:
case SCREWDRIVER: {
//lotta code
}
```

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONPart.java
DrprofLuigi: Fixed

mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
DrprofLuigi: They could, I have some that do.

mccore/src/main/java/minecrafttransportsimulator/items/instances/ItemItem.java
DrprofLuigi: The idea here was to leave it open for expansion in the future, so the screwdriver could have other uses. But they could be combined if you don't see that happening.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
```diff
@@ -255,11 +255,17 @@ protected void updateEncompassingBox() {
             return;
         }
 
-        //If we are holding a wrench, and the part has children, don't add the interaction boxes.  We can't wrench those parts.
-        //The only exceptions are parts that have permanent-default parts on them. or if they specifically don't block subpart removal.  These can be wrenched.
-        //Again, this only applies on clients for that client player.
-        if (world.isClient() && InterfaceManager.clientInterface.getClientPlayer().isHoldingItemType(ItemComponentType.WRENCH)) {
-            for (APart childPart : parts) {
+        //If we are holding a screwdriver or wrench, run these checks to remove hitboxes if needed.
+        if (world.isClient() && (InterfaceManager.clientInterface.getClientPlayer().isHoldingItemType(ItemComponentType.WRENCH) || InterfaceManager.clientInterface.getClientPlayer().isHoldingItemType(ItemComponentType.SCREWDRIVER))) {
```
DonBruce64, line 259: Good programming practice says that if you run a function twice in the same block, you should store the result.
Please add
 `boolean isHoldingWrench = InterfaceManager.clientInterface.getClientPlayer().isHoldingItemType(ItemComponentType.WRENCH)`

And the same for screwdriver, and use those booleans.  This code runs every tick, so parsing the player's held item multiple times will generate lag.  Even more speedy, one should probably get the player themselves as `player` from the first methods.

mccore/src/main/java/minecrafttransportsimulator/items/instances/ItemItem.java
DonBruce64: I would say let's code for today.  what I don't like is having two code bits that do the same thing.  That's a recipe for getting one out of whack with the other when changing one.  The paint gun vis-a-vis the crafting GUI is a perfect example.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
DrprofLuigi: Done

mccore/src/main/java/minecrafttransportsimulator/items/instances/ItemItem.java
DonBruce64: Conversation not resolved.  Request code today be combined.  Can talk about splitting at a later time.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/items/instances/ItemItem.java
DrprofLuigi: Done, sorry I missed your reply

mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
DrprofLuigi: For some reason doing this causes a crash on startup.

```
    	boolean isHoldingWrench = InterfaceManager.clientInterface.getClientPlayer().isHoldingItemType(ItemComponentType.WRENCH);
    	boolean isHoldingScrewdriver = InterfaceManager.clientInterface.getClientPlayer().isHoldingItemType(ItemComponentType.SCREWDRIVER);

        //If we are holding a screwdriver or wrench, run these checks to remove hitboxes if needed.
        if (world.isClient() && (isHoldingWrench || isHoldingScrewdriver)) {
            //If we are holding a wrench and the part requires a screwdriver, remove interaction boxes so they don't get in the way and vice versa.
            if ((isHoldingWrench && definition.generic.mustBeRemovedByScrewdriver) || (isHoldingScrewdriver && !definition.generic.mustBeRemovedByScrewdriver)) {
                allInteractionBoxes.removeAll(interactionBoxes);
                return;
            }
            //If we are holding a wrench or screwdriver, and the part has children, don't add the interaction boxes.  We can't wrench those parts.
            //The only exceptions are parts that have permanent-default parts on them. or if they specifically don't block subpart removal.  These can be removed.
            //Again, this only applies on clients for that client player.
        	for (APart childPart : parts) {
                if (!childPart.isPermanent && !childPart.placementDefinition.allowParentRemoval) {
                    allInteractionBoxes.removeAll(interactionBoxes);
                    return;
                }
            }
        }
```

But if I bypass the `isHoldingWrench` and `isHoldingScrewdriver` initialization, setting them to true, or embedding them within the checks it works fine.

DonBruce64: Yeah, you're not allowed to check the client interface on the server, so it won't work the first way.  You first should check if the world is client-side, then do the code to get the wrench or screwdriver state, and check that.  So do

if(isClient){
  bool hasScrewdriveR = condition1
  bool hasWrench = condition2
  if(hasScrewdriver || hasWrench){
      do code
  }
}

DrprofLuigi: Done! Hopefully this can still be merged. If not, I'll just redo it.

DonBruce64: The only thing that Git doesn't like is the ItemItem file, because I fixed the GUIs not opening right, and that mucks with the wrench-item code.  So if you can re-check-out the file on your end, then apply your changes on-top, it should merge.

PR #1406, opened 2022-11-03
Added ground_isspare variable

DrprofLuigi:
it is true when this vehicle is a spare. Useful when you don't want a hubcap on your spare tire.


DonBruce64:
Did you add this to the wiki?

DrprofLuigi:
I'll add it to the wiki.

PR #1409, opened 2022-11-05
sound config

muwwy:
Adding new features to the sound config:
- Randomization of sounds
- Setting the range for sound playback
- Doppler pitch effect switch.

Example:
"sounds": [
      {
        "name": "id:shot",
        "activeAnimations": [...],
        "forceSound": true,
        "minDistance": 2,
        "maxDistance": 32,
        "dopplerPitch": true,
        "randomSounds":["id:shot2","id:shot3","id:shot4"]
      }
    ]

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
@@ -580,7 +580,13 @@ public abstract class AEntityD_Definable<JSONDefinition extends AJSONMultiModelP
                         }
                     }
                     if (!isSoundPlaying) {
-                        InterfaceManager.soundInterface.playQuickSound(new SoundInstance(this, soundDef));
+                        //We get the distance from the sound to the player
+                        double distance = this.position.distanceTo(InterfaceManager.clientInterface.getClientPlayer().getPosition());
+
+                        //Checking the distance in the range of minimum and maximum values
+                        if (distance <= soundDef.maxDistance && distance >= soundDef.minDistance) {
+                            InterfaceManager.soundInterface.playQuickSound(new SoundInstance(this, soundDef));
+                        }
                     }
                 } else {
                     if (soundDef.looping) {
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
@@ -584,7 +584,7 @@ public abstract class AEntityD_Definable<JSONDefinition extends AJSONMultiModelP
                         double distance = this.position.distanceTo(InterfaceManager.clientInterface.getClientPlayer().getPosition());
 
                         //Checking the distance in the range of minimum and maximum values
-                        if ((soundDef.maxDistance != 0) && (distance <= soundDef.maxDistance && distance >= soundDef.minDistance)) {
+                        if (distance <= soundDef.maxDistance && distance >= soundDef.minDistance) {
                             InterfaceManager.soundInterface.playQuickSound(new SoundInstance(this, soundDef));
                         }
                     }
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
@@ -584,7 +584,7 @@ public abstract class AEntityD_Definable<JSONDefinition extends AJSONMultiModelP
                         double distance = this.position.distanceTo(InterfaceManager.clientInterface.getClientPlayer().getPosition());
 
                         //Checking the distance in the range of minimum and maximum values
-                        if (distance <= soundDef.maxDistance && distance >= soundDef.minDistance) {
+                        if ((soundDef.maxDistance != 0) && (distance <= soundDef.maxDistance && distance >= soundDef.minDistance)) {
                             InterfaceManager.soundInterface.playQuickSound(new SoundInstance(this, soundDef));
                         }
                     }
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
@@ -580,7 +580,7 @@ public abstract class AEntityD_Definable<JSONDefinition extends AJSONMultiModelP
                         }
                     }
                     if (!isSoundPlaying) {
-                        //We get the distance from the sound to the player
+                        //We get the distance from the sound to the player using Point3D (this may not be accurate)
                         double distance = this.position.distanceTo(InterfaceManager.clientInterface.getClientPlayer().getPosition());
 
                         //Checking the distance in the range of minimum and maximum values
--- a/mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONSound.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONSound.java
@@ -49,9 +49,9 @@ public class JSONSound {
     public List<String> soundVariations;
 
     @JSONDescription("Minimum distance for sound playback.")
-    public double minDistance;
+    public double minDistance = 0.0;
 
     @JSONDescription("Maximum distance for sound playback.")
-    public double maxDistance;
+    public double maxDistance = 64.0;
 
 }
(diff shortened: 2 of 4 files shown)


mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONSound.java
```diff
@@ -41,4 +41,17 @@ public class JSONSound {
 
     @JSONDescription("An entry of x, y, and z coordinates that define the center point of where this sound will be played relative to the center of the object.  May be omitted if you just want the sound to play at the center.")
     public Point3D pos;
+
+    @JSONDescription("Just min distance.")
+    public double minDistance = 0.0;
+
+    @JSONDescription("Just max distance.")
+    public double maxDistance = 64.0;
```
DonBruce64, line 49: Does this match vanilla sound distances?  It should, as this looks to overwrite JSONs that do not define these parameters.  Also, please verify that these defaults are applied when exporting JSONs and when parsing them.  To my recollection, variables here are overwitten with their defaults (0, false) when parsing JSON.  However, that may not be the case.

mcinterfaceforge1122/src/main/java/mcinterface1122/InterfaceSound.java
```diff
@@ -214,6 +208,18 @@ public static void update() {
...
     public void playQuickSound(SoundInstance sound) {
         if (AL.isCreated() && sourceGetFailures < 10) {
+
+            if (sound.soundDef.randomSounds != null && !sound.soundDef.randomSounds.isEmpty()) {
+                //Copying a random sounds list and adding the main one
+                List<String> sounds = new ArrayList<>(sound.soundDef.randomSounds);
+                if (sound.soundDef.name != null && !sound.soundDef.name.equals(""))
+                    sounds.add(sound.soundDef.name);
+                //Mixing sounds
+                Collections.shuffle(sounds);
```
DonBruce64, line 218: One shouldn't need to duplicate, shuffle, and get the first index of the list.  Rather, just call Math.Random to get a random number, and multiply it by the total number of random sounds.  Then floor the number to remove the decimal, and use that number to get the index in the random sound list.  Far quicker in execution.

mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONSound.java
```diff
@@ -41,4 +41,17 @@ public class JSONSound {
...
     public Point3D pos;
+
+    @JSONDescription("Just min distance.")
+    public double minDistance = 0.0;
+
+    @JSONDescription("Just max distance.")
+    public double maxDistance = 64.0;
+
+    @JSONDescription("Doppler pitch effect switch.")
+    public boolean dopplerPitch = true;
```
DonBruce64, line 52: Normally, for IV things, we go with a "change this to change the default behavior" approach.  It makes it easier for Pack Authors to understand updates.  For this, I ask that it is changed to "blockDoppler" and inverted logic.  As currently, all sounds have doppler.  So to ensure that PAs don't have to change their JSONs, the new parameter should have all sounds have doppler.

mccore/src/main/java/minecrafttransportsimulator/sound/SoundInstance.java
```diff
@@ -16,13 +16,14 @@
  * @author don_bruce
  */
 public class SoundInstance {
+
     public final AEntityB_Existing entity;
-    public final String soundName;
```
DonBruce64, line 20: This should still be final as it should not be able to be modified after starting the sound (that will cause the sound-system to be de-synced).  Please move the code to choose a random sound into the constructor here.

DonBruce64 (changes requested):

DonBruce64:
Good first PR, but a few comments on logic.  I can for sure see some good use out of this.

dldev32:
I will update the wiki article about this.

Ilya-torshov:
Doppler effect should be opt-out instead of opt-in by the way

muwwy (commented):
2.0

DonBruce64:
> Doppler effect should be opt-out instead of opt-in by the way

I mentioned that in the change requests.  I gotchu mate.

mcinterfaceforge1122/src/main/java/mcinterface1122/InterfaceSound.java
```diff
@@ -152,7 +145,7 @@ public static void update() {
 
                     //If the sound is looping, and the player isn't riding the source, calculate doppler pitch effect.
                     //Otherwise, set pitch as normal.
-                    if (sound.soundDef != null && sound.soundDef.looping && !sound.entity.equals(player.getEntityRiding())) {
+                    if (sound.soundDef != null && sound.soundDef.looping && !sound.entity.equals(player.getEntityRiding()) && !sound.soundDef.blockDoppler) {
```
DonBruce64, line 148: You'll want to check for the doppeler first (well, after making sure soundDef is not null).  This is because it won't do you any good to check looking or what the player is riding if the sound doesn't have doppler to begin with.  That's just extra code to run.

mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
```diff
@@ -580,7 +580,13 @@ public void updateSounds(float partialTicks) {
                         }
                     }
                     if (!isSoundPlaying) {
-                        InterfaceManager.soundInterface.playQuickSound(new SoundInstance(this, soundDef));
+                        //We get the distance from the sound to the player using Point3D (this may not be accurate)
+                        double distance = this.position.distanceTo(InterfaceManager.clientInterface.getClientPlayer().getPosition());
+
+                        //Checking the distance in the range of minimum and maximum values
+                        if (distance <= soundDef.maxDistance && distance >= soundDef.minDistance) {
```
DonBruce64, line 587: I noticed that this only blocks the sound from playing.  Should it also stop playing sounds outside the range?  Not required for me to approve this PR.  But if you want that functionality, it's easy enough to add since there's already logic to check if the should should be playing at all.  Just make that boolean false if you have a range and you're out if it.

DonBruce64 (changes requested):
One change requested, one food for thought that does not require changes (unless you want to make them)

mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
muwwy: I don't think it's necessary to stop sounding outside.
The logic is simple, this function is needed for the variability of sound, for example, if a gun is fired near there will be one sound, and if in the distance, then a muffled sound.

I would really like you to approve this PR, because we need this functionality.

dldev32:
Without these requirements, the PR will not be approved, do I understand correctly?

mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
DonBruce64: That is fine.  If you don't see a reason to stop the sound, then I don't mind.  Folks can submit that as a bug later if they see it as one.

PR will get approved once the first-comment about order of operations is addressed.  That's the last outstanding item.

DonBruce64:
> 

You understand incorrectly.  See my notes on that requirement.

mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
muwwy: Can this be considered a solution to our problem ? (even though it looks like a crutch)
image
image
In addition, when checking, it turned out that variables are assigned a value from JSON during export and parsing. So it really worked last time, and now.

DonBruce64: Oh no, I meant the code in InterfaceSound for the doppler effect, not that code.  There's also no way that you'll ever have 0 there, since you set that value to 64 manually if it's not present (is 0)?

muwwy: my bad, I did something incomprehensible..

muwwy: Can I leave it as it was?
Because the usual values that I set manually do not affect the values in JSON in any way
image

DonBruce64: My bad, I see what this was about.  Go ahead and put it back the way you have.

DonBruce64: Actually, it looks like it is that way.  Marking resolved.

DonBruce64:
Once you check the max sound distance, I'm good.  I want to make sure that no existing sounds play when you are more than 64 blocks away, since that will confuse PAs who had their sounds play that far previously.  Horns and sirens and engines are the ones I would check.

Alternately, you could just check if both are 0, and if so, ignore distance entirely.  TBH, that's a bit cleaner since it won't cause a new JSON parameter in every JSON.  But I'm not married to the idea so I won't force you to do that.

PR #1420, opened 2022-11-10
Made it so an empty default part string won't display an error message.

DrprofLuigi:
Useful when you want to intentionally spawn no part in `conditionalDefaultParts`

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityF_Multipart.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityF_Multipart.java
@@ -587,23 +587,24 @@ public abstract class AEntityF_Multipart<JSONDefinition extends AJSONPartProvide
      * passed-in definition is placed on this entity, not when it's being loaded from saved data.
      */
     public void addDefaultPart(String partName, int partSlot, IWrapperPlayer playerAdding, AJSONPartProvider providingDef) {
-        try {
-            String partPackID = partName.substring(0, partName.indexOf(':'));
-            String partSystemName = partName.substring(partName.indexOf(':') + 1);
-            try {
-                APart addedPart = addPartFromItem(PackParser.getItem(partPackID, partSystemName), playerAdding, InterfaceManager.coreInterface.getNewNBTWrapper(), partSlot);
-                if (addedPart != null) {
-                    //Set the default tone for the part, if it requests one and we can provide one.
-                    addedPart.updateTone(false);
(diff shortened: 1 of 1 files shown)


DonBruce64:
But why would you want to spawn no part?  You should just not include it in the conditionalDefalts?

DrprofLuigi:
```			"conditionalDefaultParts": {
				"spare_hubcap_dish"="unuparts:unuparts_part_unu_hubcap_standard_dish",
				"spare_hubcap_turbine"="unuparts:unuparts_part_unu_hubcap_standard_turbine",
				"part_isspare"="",
				"hubcap_dish"="unuparts:unuparts_part_unu_hubcap_standard_dish",
				"hubcap_turbine"="unuparts:unuparts_part_unu_hubcap_standard_turbine"
			}```
			
The vehicle as a whole has the constant to define the hubcap, but I don't want a hubcap to spawn when it is a spare tire (unless it is specifically overridden). This is also why I added the `part_isspare` variable.

File: mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityF_Multipart.java [hunk 1]
```diff
@@ -600,7 +600,9 @@ public void addDefaultPart(String partName, int partSlot, IWrapperPlayer playerA
                 playerAdding.sendPacket(new PacketPlayerChatMessage(playerAdding, "Attempted to add defaultPart: " + partPackID + ":" + partSystemName + " to: " + providingDef.packID + ":" + providingDef.systemName + " but that part doesn't exist in the pack item registry."));
             }
         } catch (IndexOutOfBoundsException e) {
-            playerAdding.sendPacket(new PacketPlayerChatMessage(playerAdding, "Could not parse defaultPart definition: " + partName + ".  Format should be \"packId:partName\""));
+            if (!partName.isEmpty()) {
```
DonBruce64, line 603: Request that rather than checking down here after the code has failed, you just don't run the addDefaultPart method at all if partName is empty.  If you know it's empty, don't even try to add the default part.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityF_Multipart.java
DrprofLuigi: Done, honestly I feel a bit dumb for doing what I did lol

PR #1423, opened 2022-11-12
Made wheel height VM-able

DrprofLuigi:
Useful for more dynamic flat tires, and swapping out different tires.


DonBruce64:
This doesn't cause any issues with part yellow boxes, or the blue wheel boxes, correct?

DrprofLuigi:
Correct, in my testing it works exactly as you would expect, and modifies the height the same as a flat tire would.

PR #1446, opened 2022-12-09
Mr.Mulles small item-texture update

SnailPerson:
- Updated texture of wrench, jumper cables, ticket, jumper-pack & key
- Updated Gunbench item-icon texture


DonBruce64 (changes requested):
Yeah, ain't no item in no core mod gonna be a higher texture than OCP vehicles.  Fix wrench 512x512 texture plz to not be crazy size.

PR #1465, opened 2022-12-20
Vairables for guided weapons

conman180:
Allows for guns to distinguish between different types of targets to lock on to. the default is it will lock how it normally does. However they can be configured to only lock planes, only lock ground vehicles, lock soft targets i.e. mobs/players.
This allows true air-to-air, or air-to-ground, etc. specific weapons.

Note: LockOnType is WIP and non functional at the moment with the exception of the default method.

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
@@ -622,7 +622,7 @@ public class PartGun extends APart {
                             }
                         }
                         //Locks only onto aircraft
-                    } else if (definition.gun.targetTypes != null && !definition.gun.targetTypes.contains(TargetTypes.ALL) && definition.gun.targetTypes.contains(TargetTypes.AIRCRAFT)) {
+                    } else if (!definition.gun.targetTypes.contains(TargetTypes.ALL) && definition.gun.targetTypes.contains(TargetTypes.AIRCRAFT)) {
                         if (entityTarget == null) {
                             engineTarget = null;
                             EntityVehicleF_Physics vehicleTargeted = world.getRaytraced(EntityVehicleF_Physics.class, controller.getPosition(), controller.getPosition().copy().add(controller.getLineOfSight(RAYTRACE_DISTANCE)), true, vehicleOn);
@@ -638,7 +638,7 @@ public class PartGun extends APart {
                             }
                         }
                         //Locks only onto ground vehicles
-                    } else if (definition.gun.targetTypes != null && !definition.gun.targetTypes.contains(TargetTypes.ALL) && definition.gun.targetTypes.contains(TargetTypes.GROUND)) {
+                    } else if (!definition.gun.targetTypes.contains(TargetTypes.ALL) && definition.gun.targetTypes.contains(TargetTypes.GROUND)) {
                         if (entityTarget == null) {
                             engineTarget = null;
                             EntityVehicleF_Physics vehicleTargeted = world.getRaytraced(EntityVehicleF_Physics.class, controller.getPosition(), controller.getPosition().copy().add(controller.getLineOfSight(RAYTRACE_DISTANCE)), true, vehicleOn);
@@ -654,7 +654,7 @@ public class PartGun extends APart {
                             }
                         }
                         //Locks only onto players or mobs. "soft targets"
-                    } else if (definition.gun.targetTypes != null && !definition.gun.targetTypes.contains(TargetTypes.ALL) && definition.gun.targetTypes.contains(TargetTypes.SOFT)) {
+                    } else if (!definition.gun.targetTypes.contains(TargetTypes.ALL) && definition.gun.targetTypes.contains(TargetTypes.SOFT)) {
                         entityTarget = world.getEntityLookingAt(controller, RAYTRACE_DISTANCE, true);
                     }
                 }
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
@@ -601,21 +601,64 @@ public class PartGun extends APart {
             //Player-controlled gun.
             //Check for a target for this gun if we have a lock-on missile.
             //Only do this once every 1/2 second.
+            //First, check if this bullet is guided
             if (loadedBullet != null && loadedBullet.definition.bullet.turnRate > 0) {
-                //Try to find the entity the controller is looking at.
-                entityTarget = world.getEntityLookingAt(controller, RAYTRACE_DISTANCE, true);
-                if (entityTarget == null) {
-                    engineTarget = null;
-                    EntityVehicleF_Physics vehicleTargeted = world.getRaytraced(EntityVehicleF_Physics.class, controller.getPosition(), controller.getPosition().copy().add(controller.getLineOfSight(RAYTRACE_DISTANCE)), true, vehicleOn);
-                    if (vehicleTargeted != null) {
-                        for (APart part : vehicleTargeted.parts) {
-                            if (part instanceof PartEngine) {
-                                engineTarget = (PartEngine) part;
-                                break;
+                //New lock on logic. Determines if it can lock a target
+                //Default old method of lokking at it to lock. Still has its uses for realistic instances.
+                if ((definition.gun.lockOnType == null || definition.gun.lockOnType == LockOnType.DEFAULT) ) {
+                    //check for valid target type. By default it lock any entity.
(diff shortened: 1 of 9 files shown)


mccore/src/main/java/minecrafttransportsimulator/packloading/LegacyCompatSystem.java
```diff
@@ -764,6 +765,12 @@ private static void performPartLegacyCompats(JSONPart definition) {
                 muzzleGroup.muzzles.add(muzzle);
                 definition.gun.muzzleGroups.add(muzzleGroup);
             }
+            if (definition.gun.lockOnType == null) {
```
DonBruce64, line 768: For this, you should check for turnRate != 0 too before assigning a lock-on-type.  You don't want to be assigning lock-ons to guns that don't do that, it could confuse pack makers.

mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONPart.java
```diff
@@ -353,6 +353,12 @@ public static class JSONPartSeat {
     }
 
     public static class JSONPartGun {
+        @JSONDescription("How a gun that has guided bullets determines if it has a lock.")
+        public LockOnType lockOnType;
+
+        @JSONDescription("A list of target types for a guided weapon. Aloows fo locking one type of thing, but not another.")
```
DonBruce64, line 359: spelling

mccore/src/main/java/minecrafttransportsimulator/packloading/LegacyCompatSystem.java
```diff
@@ -764,6 +765,12 @@ private static void performPartLegacyCompats(JSONPart definition) {
...
+                definition.gun.lockOnType = JSONPart.LockOnType.DEFAULT;
+            }
+            if (definition.gun.targetTypes == null) {
```
DonBruce64, line 771: Same as above.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -601,21 +601,64 @@ private void handleControl(IWrapperEntity controller) {
...
-                    engineTarget = null;
-                    EntityVehicleF_Physics vehicleTargeted = world.getRaytraced(EntityVehicleF_Physics.class, controller.getPosition(), controller.getPosition().copy().add(controller.getLineOfSight(RAYTRACE_DISTANCE)), true, vehicleOn);
-                    if (vehicleTargeted != null) {
-                        for (APart part : vehicleTargeted.parts) {
-                            if (part instanceof PartEngine) {
-                                engineTarget = (PartEngine) part;
-                                break;
+                //New lock on logic. Determines if it can lock a target
+                //Default old method of lokking at it to lock. Still has its uses for realistic instances.
+                if ((definition.gun.lockOnType == null || definition.gun.lockOnType == LockOnType.DEFAULT) ) {
```
DonBruce64, line 608: You shouldn't have to check if the lock-on type is null, since if it is null, you set it to DEFAULT in LegacyCompatSystem.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -601,21 +601,64 @@ private void handleControl(IWrapperEntity controller) {
...
+                    //check for valid target type. By default it lock any entity.
+                    if (definition.gun.targetTypes == null || definition.gun.targetTypes.contains(TargetTypes.ALL)) {
```
DonBruce64, line 610: Same comment as above.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -601,21 +601,64 @@ private void handleControl(IWrapperEntity controller) {
...
+                                for (APart part : vehicleTargeted.parts) {
+                                    if (part instanceof PartEngine) {
+                                        engineTarget = (PartEngine) part;
+                                        break;
+                                    }
+                                }
                             }
                         }
+                        //Locks only onto aircraft
+                    } else if (!definition.gun.targetTypes.contains(TargetTypes.ALL) && definition.gun.targetTypes.contains(TargetTypes.AIRCRAFT)) {
```
DonBruce64, line 625: You don't need to check if targetTypes contains ALL here, since if it did, you'd go into the logic above via line 610.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -601,21 +601,64 @@ private void handleControl(IWrapperEntity controller) {
...
+                                        if (part instanceof PartEngine) {
+                                            engineTarget = (PartEngine) part;
+                                            break;
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                        //Locks only onto ground vehicles
+                    } else if (!definition.gun.targetTypes.contains(TargetTypes.ALL) && definition.gun.targetTypes.contains(TargetTypes.GROUND)) {
```
DonBruce64, line 641: I'm thinking perhaps you want to have 4 individual targeting types?  ALL, AIRCRAFT, GROUND, and SOFT. In this case, you wouldn't want a list of possible types, but instead choose one target type.  If you want something that only hits hard targets like AIRCRAFT and GROUND, but not SOFT, perhaps also add HARD to encompass that?  If you do that, then you can change targetTypes to targetType, make it just a TargetType rather than a list, and use a switch statement here, which is far easier and cleaner to follow, and is the more-correct way to use enums like you've made.  Totally get what you're going for here though and totally approve.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -601,21 +601,64 @@ private void handleControl(IWrapperEntity controller) {
...
+                //Default old method of lokking at it to lock. Still has its uses for realistic instances.
+                if ((definition.gun.lockOnType == null || definition.gun.lockOnType == LockOnType.DEFAULT) ) {
+                    //check for valid target type. By default it lock any entity.
+                    if (definition.gun.targetTypes == null || definition.gun.targetTypes.contains(TargetTypes.ALL)) {
+                        entityTarget = world.getEntityLookingAt(controller, RAYTRACE_DISTANCE, true);
+                        if (entityTarget == null) {
+                            engineTarget = null;
+                            EntityVehicleF_Physics vehicleTargeted = world.getRaytraced(EntityVehicleF_Physics.class, controller.getPosition(), controller.getPosition().copy().add(controller.getLineOfSight(RAYTRACE_DISTANCE)), true, vehicleOn);
+                            if (vehicleTargeted != null) {
+                                for (APart part : vehicleTargeted.parts) {
```
DonBruce64, line 616: I know that this was how the origional code was written, but it should be vehicleTargeted.allParts, not vehicleTargeted.parts.  This is because parts only gets the parts on the vehicle, not sub-parts.  So if you have a wing-mount with an engine on it, that won't be seen by the gun.  If there's 0 chance of parts having the only engine on a vehicle, please ignore this comment, as checking on the vehicle is far quicker and since this code runs every tick, any speed-ups are important.  Might be a question for the war-focused PAs?

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
conman180: I have this one as a list because it leaves the ability for PAs to have that choice of combinations. I get what you're trying to say here with only having one. As for switches this must be why I couldn't get one to work. I'll change this and see what it does.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
conman180: I'd pose the question to PAs. I've never seen a part on a vehicle with an engine sub-part yet.

mccore/src/main/java/minecrafttransportsimulator/packloading/LegacyCompatSystem.java
DonBruce64: I was mistaken about this, turnRate is on the bullet, so it doesn't apply here.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
boot2big: I haad one, I changed it anyways because I found a better method to work with it though, but I had one.

PR #1473, opened 2022-12-27
Modified canBeClicked()

DrprofLuigi:
Currently interactable variables do not block child parts. This fixes that. Now if a part is not interactable, then all the child parts will also not be interactable.

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
@@ -339,7 +339,7 @@ public abstract class APart extends AEntityF_Multipart<JSONPart> {
 
     @Override
     public boolean canBeClicked() {
-        return entityOn.isVariableListTrue(placementDefinition.interactableVariables);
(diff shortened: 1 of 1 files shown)


mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
```diff
@@ -339,7 +339,7 @@ public boolean shouldSavePosition() {
 
     @Override
     public boolean canBeClicked() {
-        return entityOn.isVariableListTrue(placementDefinition.interactableVariables);
+        return entityOn.isVariableListTrue(placementDefinition.interactableVariables) ? entityOn.canBeClicked() : false;
```
DonBruce64, line 342: The techincally correct way to write this is as an AND statement.  `entityOn.isVariableListTrue(placementDefinition.interactableVariables) && entityOn.canBeClicked()`.  Makes it clearer saying "variable list must be true, and entity on must able to be clicked."

DonBruce64 (changes requested):

DonBruce64 (approved):

PR #1501, opened 2023-01-09
Fix Tracking bug

conman180:
Added .copy() to function as old function was moving the target's engine part out to the predicted impact point rather than the bullet just aiming there.
Bug found by VEB's Wartich.

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
@@ -164,10 +164,6 @@ public class EntityBullet extends AEntityD_Definable<JSONBullet> {
                     targetPosition = null;
                 }
             } else if (engineTargeted != null) {
-                Point3D predictedCollisionPoint = new Point3D();
-                double ticksToTarget = targetPosition.distanceTo(position) / (velocity / 20D / 10D);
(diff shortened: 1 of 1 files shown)


DonBruce64:
Are you sure this is what fixes it?  Looking at the old code, I see no modification being made to the engine's position.  You set the target to the value of the engine's position, and modify the target vector, but you never modify the engine's position vector.

conman180:
Just figured out this did not solve the issue. will commit changes once confirmed solution is found.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
```diff
@@ -181,7 +177,9 @@ public void update() {
                 if (targetVector == null) {
                     targetVector = new Point3D();
                 }
-                targetVector.set(targetPosition).subtract(position).reOrigin(orientation).getAngles(true);
+                double ticksToTarget = targetPosition.distanceTo(position) / (velocity / 20D / 10D);
+                targetVector.set(targetPosition.copy().addScaled(engineTargeted.vehicleOn.motion, (engineTargeted.vehicleOn.speedFactor/ 20D/10D) * ticksToTarget)).subtract(position).reOrigin(orientation).getAngles(true);
```
DonBruce64, line 181: You don't need to copy the targetPosition here.  I know why you're doing it: because if you don't, and you addScaled to it, then you modify that vector.  But what you should do is set the targetVector to targetPosition, then do the modifications to targetVector itself.

DonBruce64 (changes requested):

conman180:
That should fix the lead calculation. Or are there any other changes you request?

PR #1504, opened 2023-01-12
[PR] New damaged engine code

boot2big:
Engines will last a little longer, since 250 hours still doesn't seem to be enough IMO.
I had previously attempted to get PAs together to discuss this, but it'd have been easier to wrangle together a bunch of stray cats and ask _them_ instead. Pro'lly a little more civilized as well.

Additionally gets rid of booger code that was created directly because of my previous engine PR, which should also bring back a surprise feature that adds to immersion without being a PITA to end users. I'll let you figure it out ingame...

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
@@ -425,8 +425,8 @@ public class PartEngine extends APart {
                         }
 
                         //If the engine has high hours, give a chance for a backfire.
-                        if (hours > 250 && !world.isClient()) {
-                            if (Math.random() < (hours / 2) / (250 + (10000 - hours)) * (currentMaxSafeRPM / (rpm + currentMaxSafeRPM / 1.5))) {
+                        if (hours >= 500 && !world.isClient()) {
+                            if (Math.random() < (hours / 3) / (500 + (10000 - hours)) * (currentMaxSafeRPM / (rpm + currentMaxSafeRPM / 1.5))) {
                                 backfireEngine();
                                 InterfaceManager.packetInterface.sendToAllClients(new PacketPartEngine(this, Signal.BACKFIRE));
                             }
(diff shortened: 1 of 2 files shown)


mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
```diff
@@ -544,11 +544,8 @@ public void update() {
                     if (wheelFriction > 0) {
                         double desiredRPM = lowestWheelVelocity * 1200F * currentGearRatio * vehicleOn.currentAxleRatio;
                         rpm += (desiredRPM - rpm) / definition.engine.revResistance;
-                        if (rpm < currentIdleRPM && running && backfireCooldown <= 0) {//Checks if we're backfiring and sets lugging rpm to stall rpm, otherwise sets lug rpm to idle
-                            rpm = currentIdleRPM;
-                        } else if (rpm < definition.engine.stallRPM && running) {
-                            rpm = definition.engine.stallRPM;
-                            backfireCooldown -= 1;
+                        if (rpm < (currentIdleRPM - ((currentIdleRPM - definition.engine.stallRPM) * 0.5)) && running) {
```
DonBruce64, line 547: This here, will it cause engines with hours to stall out and die if they get too low RPM?  It shouldn't....  But it looks like it will.

boot2big: That... was the idea? That was even the point of the original booger code! If they're in gear and sputter too much they stall and die, but if they're not sputtering then it runs as normal.

boot2big: To clarify, what this here does is make it so _(every)_ engine idles between their idle & stall rpm while in gear. On electrics this has no effect because... 0 minus 0 multiplied by 0.5 (halving without dividing) equals... 0. On non-damaged combustion engines this causes it to idle slightly lower, but won't do much else (and it specifically waits for the engine to get to that RPM before setting it to that as well), and on damaged engines it's basically just a better version of the previous code since, you _are_ close enough to stallRPM to sputter out, but it's no longer a guarantee that you'll sputter out either. So if anything, this code's an improvement on more than one end!

boot2big:
To recap:
Engines will begin to malfunction at 500 hours, at a low frequency. This still ramps up as more hours are accumulated.
At lower engine RPMs, the gremlins prefer to sputter out the engine more, but they do not have as much oomph and so it takes sequential backfires through several ticks in a row to risk stalling the engine out completely. 
At higher engine RPMs, the gremlins take offense to this and hit harder rather than trying to rapid-fire sputter the engine, which brings a larger RPM penalty.
Combined, this results in a system that does pose the user-risk of an engine staying near lower RPMs, but not immediately sputtering out.

Consequently the way engines idle in gear had to be changed, since this was crappy both from a gameplay and code perspective. 
Previously they would just switch from idleRPM to stallRPM for a few ticks after each backfire, which did "work" but would risk an engine stalling out in just two sputters, possibly even one if idleRPM and stallRPM were close enough together.
Now, MTS will simply find a value between the idleRPM and stallRPM to "lug" the engine at, which simultaneously also fixes a bug that the previous system introduced/
By killing two birds with one stone, engines are simultaneously less likely to instantly shut off while in gear _and_ ground vehicles can now idle accelerate once again, an example of simple yet intuitive immersiveness. No extra complications or useless parts, just a simple "Put 'er in gear to get movin', and push on the gas to go faster!" system.

PR #1512, opened 2023-01-15
Add missing import from prior PR

conman180:


mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
```diff
@@ -177,6 +178,7 @@ public void update() {
                 if (targetVector == null) {
                     targetVector = new Point3D();
                 }
+                
                 double ticksToTarget = targetPosition.distanceTo(position) / (velocity / 20D / 10D);
                 if ((gun.definition.gun.targetType == JSONPart.TargetType.ALL && engineTargeted != null) || gun.definition.gun.targetType == JSONPart.TargetType.AIRCRAFT || gun.definition.gun.targetType == JSONPart.TargetType.GROUND) {
```
DonBruce64, line 183: FYI: See how we only use BulletType in the code and import JSONBullet.BulletType?  Should be the same for importing JSONPart.TargetType and only using TargetType.

DonBruce64 (changes requested):

PR #1516, opened 2023-01-19
Add some new guidance mechanics

conman180:
Adds Semi-Active Guidance for missiles, which means the gun firing must maintain its lock on a target for the bullet to guide to it. Once lock is broken off, guidance stops.
Adds Laser Guidance in the best way it can right now as BlockHitResults are limited to 200 blocks before the raytrace returns null.
Adds a LegacyCompat for guidanceType

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
@@ -88,7 +88,7 @@ public class EntityBullet extends AEntityD_Definable<JSONBullet> {
     /**
      * Positional target.
      **/
-    private EntityBullet(Point3D position, Point3D motion, RotationMatrix orientation, PartGun gun, Point3D targetPosition) {
+    public EntityBullet(Point3D position, Point3D motion, RotationMatrix orientation, PartGun gun, Point3D targetPosition) {
         this(position, motion, orientation, gun);
         this.targetPosition = targetPosition;
     }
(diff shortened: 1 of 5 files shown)


conman180:
Countermeasures are planned for the future but after dealing with the gun and bullet code a bit, it might require an overhaul of the wholes system. I want countermeasures to be realistic, but for that I have to make the lock mechanics work first. E.I> this adding semi-active guidance as a start. currently guns lock engines and MC entities only. Id have to re work it to be, heat signatures(running engines, flare type bullets, block on fire) and be able for a bullet not assigned a target to see said things after it's been fired. etc. On the whole i wanna take small steps and get prerequisite things working first

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
```diff
@@ -153,38 +153,77 @@ public void update() {
...
-
             if (targetPosition != null) {
                 //Get the angular delta between us and our target, in our local orientation coordinates.
                 if (targetVector == null) {
                     targetVector = new Point3D();
                 }
+                //Lead the target. Otherwise, just turn to face it
                 double ticksToTarget = targetPosition.distanceTo(position) / (velocity / 20D / 10D);
-                if ((gun.definition.gun.targetType == TargetType.ALL && engineTargeted != null) || gun.definition.gun.targetType == TargetType.AIRCRAFT || gun.definition.gun.targetType == TargetType.GROUND) {
+                if ((gun.definition.gun.targetType == TargetType.ALL && engineTargeted != null) || (gun.definition.gun.targetType == TargetType.AIRCRAFT  && engineTargeted != null) || (gun.definition.gun.targetType == TargetType.GROUND  && engineTargeted != null)) {
```
DonBruce64, line 220: So, a bit of redundancy here, for all checks, you're looking of engineTargeted != null, so simpler way of writing this is: `(engineTargeted != null && (gun.definition.gun.targetType == TargetType.ALL  || gun.definition.gun.targetType == TargetType.AIRCRAFT || gun.definition.gun.targetType == TargetType.GROUND)`

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
```diff
@@ -153,38 +153,77 @@ public void update() {
...
                     targetVector.set(targetPosition).addScaled(engineTargeted.vehicleOn.motion, (engineTargeted.vehicleOn.speedFactor / 20D / 10D) * ticksToTarget).subtract(position).reOrigin(orientation).getAngles(true);
-                } else if ((gun.definition.gun.targetType == TargetType.ALL && externalEntityTargeted != null) || gun.definition.gun.targetType == TargetType.SOFT) {
+                } else if ((gun.definition.gun.targetType == TargetType.ALL && externalEntityTargeted != null) || (gun.definition.gun.targetType == TargetType.SOFT && externalEntityTargeted != null)) {
```
DonBruce64, line 222: Same as above.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -659,7 +664,16 @@ private void handleControl(IWrapperEntity controller) {
                         break;
                     }
                     case MANUAL: {
-                        //No target to set, and no actions to perform.
+                        if (targetPosition == null) {
+                            targetPosition = new Point3D();
+                        }
+                        Point3D laserStart = new Point3D(controller.getPosition().x, controller.getPosition().y, controller.getPosition().z);
```
DonBruce64, line 670: Point3Ds don't need to be set with each component.  So you can just do `controller.getPosition().copy()`

DonBruce64 (changes requested):

PR #1533, opened 2023-02-12
[PR] Engine rev limiter variables + gradlew executable by default

boot2big:
Variables to report currentRevlimitRPM as well as the percentage of rpm compared to the revlimitRPM (which defaults to maxSafeRPM if revlimitRPM is -1)


boot2big:
Oh yeah. Looks like my repo is a bit broken too. Oops.

Will be fixed next time I make a PR, I swear!

DonBruce64 (changes requested):
You changed the gradlew file.  Please don't do that.  It makes things not work well.

boot2big:
I only changed the permissions though, as by default it ain't executable. I can yeet that if just adding 1 to each permission will break stuff though.

DonBruce64:
Add to the wiki.

boot2big:
Is Wiki'd.

PR #1546, opened 2023-02-25
A few things cant remember all of them.

conman180:
small fixes and additions for things.


mccore/src/main/java/minecrafttransportsimulator/packloading/LegacyCompatSystem.java
```diff
@@ -201,6 +201,8 @@ private static void performVehicleLegacyCompats(JSONVehicle definition) {
             definition.general.openTop = false;
             definition.motorized.emptyMass = definition.general.emptyMass;
             definition.general.emptyMass = 0;
+            definition.general.radarRange = 512;
```
DonBruce64, line 204: You probably don't want this here.  This will only work on older JSONs.  If someone has a newer JSON where emptyMass is defined in the motorized section per standards, it won't auto-add your radar parameters.  You will want to check if they are 0 outside this if statement.

mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
```diff
@@ -210,7 +203,11 @@ public int compare(AEntityB_Existing o1, AEntityB_Existing o2) {
                 groundersOnRadar.clear();
             }
             for (EntityVehicleF_Physics vehicle : allVehicles) {
-                if (!vehicle.outOfHealth && vehicle != this) {
+                Point3D searchVector = new Point3D(0, 0, definition.general.radarRange).rotate(orientation);
```
DonBruce64, line 206: Since you're iterating over all vehicles, you want to make searchVector once, outside the loop, then set the value of it inside the loop.  Far faster than creating a new object (or two, given there's another Point3D in  here) each vehicle..

mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
```diff
@@ -210,7 +203,11 @@ public int compare(AEntityB_Existing o1, AEntityB_Existing o2) {
...
+                Point3D LOSVector = new Point3D().set(vehicle.position).subtract(position).normalize();
+                double coneAngle = definition.general.radarWidth;
+                double angle = Math.abs(Math.toDegrees(Math.acos(searchVector.normalize().dotProduct(LOSVector, false))));
+                if (!vehicle.outOfHealth && vehicle != this && (angle < coneAngle && vehicle.position.distanceTo(position) < definition.general.radarRange)) {
```
DonBruce64, line 210: Rather than doing distanceTo, there's a function isDistanceToCloserThan for Point3D.  That function is more efficient as it doesn't calculate square roots.  MC does this for most, if not all, of its distance checks, since square roots are CPU intensive.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/packloading/LegacyCompatSystem.java
conman180: I put this in the general section that way AA turrets and the like have the ability to make use of the radar. To my knowledge I didnt do anything to emptyMass.

conman180: retracted comment. understand what you mean now.

conman180: commit on this removes legacy compat as it auto-reverts to zero and doesnt crash if not included in the json.

PR #1615, opened 2023-04-27
Fixed IE mappings

DrprofLuigi:


DrprofLuigi:
Hold off on pulling this one in, there are some issues with the vanilla list. The first _only_ had 'flattened' data (it omits anything that just needs the `:0` removed), but I 'fixed' it with an incomplete list of items...

DonBruce64:
If the item isn't flattened, then it doesn't need to be in the list.  Essentially, the system will auto-remove the 0 if it can't find a match.

DrprofLuigi:
Alright. I updated the mappings to include all normally obtainable items (so it omits things like the 'wood' blocks (log with bark on all sides) and things like potions and eggs that use metadata.

I can remove items that have `:0` if you want.

DonBruce64:
0 elimination would be ideal, as it'll cut down on file size and redundancy.

DrprofLuigi:
Done, should be all good to go.

PR #1621, opened 2023-05-03
Allowed manual defintion of the groundDeviceBox

### LINKED ISSUE #1484: Sometimes front and rear wheels are combined into the same wheel box...
Here is a picture

The top UNU Baron has the blue wheel box in the middle, while the bottom one has the boxes on both wheels.

This is likely due to how the suspension is animated, the wheel sometimes pivots in front of z=0.

I would suggest one of two solutions:
The wheel calculation is not dynamic, so what wheel is in which 'region' is determined by placement only. The other solution would be to have a manual override to hard-define which 'region' the wheel should be tied to..

DrprofLuigi:
Fixes #1484

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/baseclasses/VehicleGroundDeviceBox.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/baseclasses/VehicleGroundDeviceBox.java
@@ -101,8 +101,9 @@ public class VehicleGroundDeviceBox {
             if (part instanceof PartGroundDevice) {
                 if (!part.isSpare) {
                     PartGroundDevice ground = (PartGroundDevice) part;
-                    if (isFront && ground.wheelbasePoint.z > 0) {
-                        if (isLeft && ground.wheelbasePoint.x >= 0) {
+                    int groundDeviceBoxNumber = part.placementDefinition.groundDeviceBoxNumber;
+                    if (isFront && ((ground.wheelbasePoint.z > 0 && groundDeviceBoxNumber == 0) || 3 <= groundDeviceBoxNumber)) {
+                        if (isLeft && ((ground.wheelbasePoint.x >= 0 && groundDeviceBoxNumber == 0) || groundDeviceBoxNumber == 4)) {
                             groundDevices.add(ground);
                             totalClimbHeight += ground.definition.ground.climbHeight;
                             if (ground.definition.ground.isWheel || ground.definition.ground.isTread) {
@@ -111,7 +112,7 @@ public class VehicleGroundDeviceBox {
                             if (ground.definition.ground.canFloat) {
                                 liquidDevices.add(ground);
                             }
-                        } else if (!isLeft && ground.wheelbasePoint.x <= 0) {
+                        } else if (!isLeft && ((ground.wheelbasePoint.x <= 0 && groundDeviceBoxNumber == 0) || groundDeviceBoxNumber == 3)) {
                             groundDevices.add(ground);
                             totalClimbHeight += ground.definition.ground.climbHeight;
                             if (ground.definition.ground.isWheel || ground.definition.ground.isTread) {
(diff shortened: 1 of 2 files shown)


DrprofLuigi:
This adds a part slot definition of `groundDeviceBoxNumber` where you can hard-define the corner the wheel is assigned to.

mccore/src/main/java/minecrafttransportsimulator/baseclasses/VehicleGroundDeviceBox.java
```diff
@@ -101,8 +101,9 @@ public void updateMembers() {
             if (part instanceof PartGroundDevice) {
                 if (!part.isSpare) {
                     PartGroundDevice ground = (PartGroundDevice) part;
-                    if (isFront && ground.wheelbasePoint.z > 0) {
-                        if (isLeft && ground.wheelbasePoint.x >= 0) {
+                    int groundDeviceBoxNumber = part.placementDefinition.groundDeviceBoxNumber;
+                    if (isFront && ((ground.wheelbasePoint.z > 0 && groundDeviceBoxNumber == 0) || 3 <= groundDeviceBoxNumber)) {
```
DonBruce64, line 105: I'm thinking this line isn't right, since it has a <= in it rather than a ==.

mccore/src/main/java/minecrafttransportsimulator/baseclasses/VehicleGroundDeviceBox.java
```diff
@@ -121,8 +122,8 @@ public void updateMembers() {
                                 liquidDevices.add(ground);
                             }
                         }
-                    } else if (!isFront && ground.wheelbasePoint.z <= 0) {
-                        if (isLeft && ground.wheelbasePoint.x >= 0) {
+                    } else if (!isFront && ((ground.wheelbasePoint.z <= 0 && groundDeviceBoxNumber == 0) || (1 <= groundDeviceBoxNumber && groundDeviceBoxNumber <= 2))) {
```
DonBruce64, line 125: (1 <= groundDeviceBoxNumber && groundDeviceBoxNumber <= 2)
This just checks if the GDVB is 1 or 2, so a cleaner way of doing this is:
(1 == groundDeviceBoxNumber || groundDeviceBoxNumber == 2)
Same with others.

mccore/src/main/java/minecrafttransportsimulator/baseclasses/VehicleGroundDeviceBox.java
```diff
@@ -131,7 +132,7 @@ public void updateMembers() {
                             if (ground.definition.ground.canFloat) {
                                 liquidDevices.add(ground);
                             }
-                        } else if (!isLeft && ground.wheelbasePoint.x <= 0) {
+                        } else if (!isLeft && ((ground.wheelbasePoint.x <= 0 && groundDeviceBoxNumber == 0) || groundDeviceBoxNumber == 2)) {
```
DonBruce64, line 135: As a general thought though, have you considered an enum for this?  Where you make this a string-based paramter like frontLeft, frontRight, rearLeft, rearRight?  With an enum, you can also add simple functions such as groundDeviceBox.isRight() or .isFront().  It's clearer to the PA since they don't need to remember which number goes to what, and cuts down on these odd equalities.  See ABlockBase.Axis for more details.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/baseclasses/VehicleGroundDeviceBox.java
DrprofLuigi: I modified it to use a string definition. I'm not sure how to use enum in this context.

DonBruce64: @DrprofLuigi Changed to use enums.  You'll wanna test my logic to be sure it meets your needs.  I added center-wheels, since that's required for things like the Trio and the Auroras for their front and rear wheels respectively which are both left and right wheels.

DrprofLuigi:
Is there anything else that needs to be done on this (i.e. do I need to go in and to the internal definitions to use the enum values)?

DonBruce64:
I've changed all the code to work in both files, but you should probably test it to make sure it does what you want it to still and I didn't screw it up.

DrprofLuigi:
I can confirm it works. I'll update the wiki when this PR is merged.

DonBruce64:
Merged and ready for wiki entry.

PR #1634, opened 2023-05-17
Round out `acceleration` variable from -+0.025 to 0 in order to mitigate negative-positive jumping

boot2big:
This freakin' thing has made me lose my mind for the LAST time.
No longer will I concede with making the translation or rotation guided by this variable smaller, nor hath I the patience to find just the right amount to dampen this variable by.
This god-forsaken value jumping padded by more 0's than π itself ends here, `acceleration` variable! I win!! 
image

tl;dr Rounding the `acceleration` so it doesn't constantly switch between positive and negative values, which doesn't play well with most uses for this variable


boot2big:
Sorry, it was late at night when I wrote that.

PR #1640, opened 2023-05-27
Allowed for more complex VM Math

DrprofLuigi:
This PR adds trig functions as well as square rooting and division to VM math. This also adds an 'invert' boolean field to animations (that is only used in VMs for now)

In the following expressions, 'MV' is `modifiedValue` or the 'running total' of the VM. `V` is the input variable value, `[X, Y, Z]` is the axis, and `[x, y, z]` is the centerPoint.

[Redacted Translation changes, see later comment]

The trig functions utilize the 'rotation' animationType.

The output is simply:

`MV * (Xsin(V + x) Ycos(V + y) + Ztan(V + z))` (inputs in degrees)

When inverted, the trig functions are swapped to their inverses (sine becomes arcsine, etc.).

There are internal conversions in the code to keep radians in their place.

This picture shows the test values, and I tested it fairly thoroughly:

Here are the VMs that produced the above output:
```		{
			"variable": "clock",
			"animations": [
				{
					"animationType": "translation",
					"variable": "unuisbest",
					"axis": [0, 1, 0]
				},
				{
					"animationType": "inhibitor",
					"variable": "clock",
					"clampMin": 1,
					"clampMax": 360
				},
				{
					"animationType": "translation",
					"variable": "unuisbest",
					"axis": [0, 0, 1]
				}
			]
		},
		{
			"variable": "clock_sin",
			"animations": [
				{
					"animationType": "translation",
					"variable": "unuisbest",
					"axis": [0, 0, 1]
				},
				{
					"animationType": "rotation",
					"variable": "clock",
					"centerPoint": [0, 0, 0],
					"axis": [1, 0, 0]
				}
			]
		},
		{
			"variable": "clock_cos",
			"animations": [
				{
					"animationType": "translation",
					"variable": "unuisbest",
					"axis": [0, 0, 1]
				},
				{
					"animationType": "rotation",
					"variable": "clock",
					"centerPoint": [0, 0, 0],
					"axis": [0, 1, 0]
				}
			]
		},
		{
			"variable": "clock2",
			"animations": [
				{
					"animationType": "translation",
					"variable": "unuisbest",
					"axis": [0, 0.01, 0]
				},
				{
					"animationType": "inhibitor",
					"variable": "clock2",
					"clampMin": -1,
					"clampMax": 2
				},
				{
					"animationType": "translation",
					"variable": "unuisbest",
					"axis": [0, 0, -1]
				}
			]
		},
		{
			"variable": "clock2_asin",
			"animations": [
				{
					"animationType": "translation",
					"variable": "unuisbest",
					"axis": [0, 0, 1]
				},
				{
					"animationType": "rotation",
					"variable": "clock2",
					"centerPoint": [0, 0, 0],
					"axis": [1, 0, 0],
					"invert": true
				}
			]
		},
		{
			"variable": "clock2_acos",
			"animations": [
				{
					"animationType": "translation",
					"variable": "unuisbest",
					"axis": [0, 0, 1]
				},
				{
					"animationType": "rotation",
					"variable": "clock2",
					"centerPoint": [0, 0, 0],
					"axis": [0, 1, 0],
					"invert": true
				}
			]
		},
		{
			"variable": "test",
			"animations": [
				{
					"animationType": "translation",
					"variable": "unuisbest",
					"axis": [0, 0, 5]
				}
			]
		},
		{
			"variable": "test2",
			"animations": [
				{
					"animationType": "translation",
					"variable": "unuisbest",
					"axis": [0, 0, 1]
				},
				{
					"animationType": "translation",
					"variable": "test",
					"axis": [1, 0.5, 0]
				}
			]
		},
		{
			"variable": "test3",
			"animations": [
				{
					"animationType": "translation",
					"variable": "unuisbest",
					"axis": [0, 0, 1]
				},
				{
					"animationType": "translation",
					"variable": "test",
					"axis": [1, 0.3333, 0]
				}
			]
		}```

### CODE REVIEW

**Comment by DrprofLuigi** [comment]
After thinking about it for a moment, I realize the 'invert' tag for the translation animation is fairly redundant.

A cleaner way to do it would be to do the following for the translation logic.

If X != 0 and Y !=0, then the output is:

MV * (X * V^Y)

This allows for division by setting Y to -1, and allows for simple squaring and rooting. This should also keep things backwards-compatible because iirc everyone that uses VMs to multiply uses an axis of `[X, 0, 0]` This would eliminate the invert tag from being used in the translation case altogether. 

It is also unlikely that someone would want to do `MV * (X * V^0)` since that is just multiplying by a constant (which could be done easily by just multiplying by a constant-defined variable), so using Y = 0 as a backwards-compatible case shouldn't be an issue. Though I would ask the MTS discord to make sure there aren't any sloppy VM multiplications that may get messed up.

File: mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java [hunk 1]
```diff
@@ -1125,13 +1125,20 @@ private VariableModifierSwitchbox(AEntityD_Definable<?> entity, List<JSONAnimati
         @Override
         public void runTranslation(DurationDelayClock clock, float partialTicks) {
             if (clock.animation.axis.x != 0) {
-                modifiedValue *= entity.getAnimatedVariableValue(clock, clock.animation.axis.x, partialTicks);
+                modifiedValue *= Math.pow(entity.getAnimatedVariableValue(clock, clock.animation.axis.x, partialTicks), clock.animation.axis.y != 0 ? clock.animation.axis.y : 1);
```
DonBruce64, line 1128: This should be an if check.  Don't run Pow functions if you know you don't need to and will only raise to the power of 1.  Pow functions are a significant overhead and should only be ran when absolutely required.  Same for square root.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
DrprofLuigi: Done

mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONAnimationDefinition.java
```diff
@@ -35,6 +35,9 @@ public class JSONAnimationDefinition {
     @JSONDescription("If true, the absolute value of the variable will be used rather than the actual value.  Note that negative movements are still possible via a negative rotation factor or a large negative offset.")
     public boolean absolute;
 
+    @JSONDescription("Used for VMs. If true, trig functions will be inverted.")
```
DonBruce64, line 38: I noticed that you said invert was redundant.  Is that still the case?  I don't see it used in the code, but I do see it in the JSON here.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONAnimationDefinition.java
DrprofLuigi: Inverting for the translation was redundant. It is still necessary for the rotation (trig) case.

It was redundant for the translation case because I had it trigger rooting and division, but with the other exponent method, rooting and division can just be done with an exponent.

PR #1674, opened 2023-06-30
Added variable modifier for isAutomatic

boot2big:
Works similarly to isTwoHanded, forceShift, etc.
Still checks engine definition, but also allows pack authors to switch their engines between manual and automatic using a variable modifier.


DonBruce64:
Works for me.  Just make sure the variable gets asterikized in the wiki so folks know you can VM it.

boot2big:
https://github.com/DonBruce64/MinecraftTransportSimulator/wiki/Pack-Making-JSON-Parts
 I liked the part where he said "it's wiki'ing time" and he wiki'd all over this page!

PR #1678, opened 2023-07-05
Bullets & Guns Stuff

conman180:
Ignore commits for the gravity stuff as that was already merged. Don't know why it still shows up here.
First commit: makes it so active guidance bullets have that same search cone as guns do. if the target its tracking goes outside that cone, it loses lock. Also makes a check if terrain is in between bullet and target, allowing for terrain to break said lock as well. may need a little tweaking still but need your help with it.

Second commit: Adds a check for if the gun that has a lock on a target is on a vehicle, if it is, it bases the reference point for animations of of the vehicle pos rather than gun pos. this is to help animations on HUDs stay lined up on larger aircraft.


conman180:
Will add those two new json defs for bullets to the wiki once merged.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -1052,9 +1052,9 @@ public void setBulletSpawn(Point3D bulletPosition, Point3D bulletVelocity, Rotat
     public double getLockedOnDirection(){
         double direction = 0;
         if (engineTarget != null) {
-            direction = Math.toDegrees(Math.atan2(-engineTarget.position.z + position.z, -engineTarget.position.x + position.x)) + 90 + orientation.angles.y;
+            direction = Math.toDegrees(Math.atan2(-engineTarget.vehicleOn.position.z + (vehicleOn!=null ?vehicleOn.position.z : position.z), -engineTarget.vehicleOn.position.x + (vehicleOn!=null ?vehicleOn.position.x : position.x))) + 90 + orientation.angles.y;
```
DonBruce64, line 1055: You check if the vehicleOn is not null three times.  Better thing to do is make a variable up by double direction called "startPoint" or something and set that to vehicleOn.position or position and reference that variable down here.  Cleaner code.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -1119,7 +1119,7 @@ public double getRawVariableValue(String variable, float partialTicks) {
...
-                return entityTarget != null ? (-Math.toDegrees(Math.atan2(-entityTarget.getPosition().y + position.y,Math.hypot(-entityTarget.getPosition().z + position.z,-entityTarget.getPosition().x + position.x))) + orientation.angles.x) : (engineTarget != null ? (-Math.toDegrees(Math.atan2(-engineTarget.position.y + position.y,Math.hypot(-engineTarget.position.z + position.z,-engineTarget.position.x + position.x))) + orientation.angles.x) : 0);
+                return entityTarget != null ? (-Math.toDegrees(Math.atan2(-entityTarget.getPosition().y + (vehicleOn!=null ?vehicleOn.position.y : position.y),Math.hypot(-entityTarget.getPosition().z + (vehicleOn!=null ?vehicleOn.position.z : position.z),-entityTarget.getPosition().x + (vehicleOn!=null ?vehicleOn.position.x : position.x)))) + orientation.angles.x) : (engineTarget != null ? (-Math.toDegrees(Math.atan2(-engineTarget.position.y + (vehicleOn!=null ?vehicleOn.position.y : position.y),Math.hypot(-engineTarget.position.z + (vehicleOn!=null ?vehicleOn.position.z : position.z),-engineTarget.position.x + (vehicleOn!=null ?vehicleOn.position.x : position.x)))) + orientation.angles.x) : 0);
```
DonBruce64, line 1122: Same as above comment.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -1051,10 +1051,16 @@ public void setBulletSpawn(Point3D bulletPosition, Point3D bulletVelocity, Rotat
 
     public double getLockedOnDirection(){
         double direction = 0;
+        Point3D referencePos = new Point3D();
```
DonBruce64, line 1054: You don't need to make a new point here though?  You only reference the point you choose in the older version.  So `Point3D referencePos = vehicleOn != null ? vehicleOn.position : position`.  Don't create variables you don't need  if you can just reference existing ones.  `new` makes new variables.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -1087,7 +1114,15 @@ public double getLeadPointDirection(){
         return direction;
     }
 
-
+    public double getLeadAngleY() {
+        double angle = 0;
+        if(engineTarget != null) {
+            angle = (-Math.toDegrees(Math.atan2(-getLockedOnLeadPoint().y + position.y, Math.hypot(-getLockedOnLeadPoint().z + position.z, -getLockedOnLeadPoint().x + position.x))) + orientation.angles.x) - (-Math.toDegrees(Math.atan2(-engineTarget.position.y + (vehicleOn != null ? vehicleOn.position.y : position.y), Math.hypot(-engineTarget.position.z + (vehicleOn != null ? vehicleOn.position.z : position.z), -engineTarget.position.x + (vehicleOn != null ? vehicleOn.position.x : position.x)))) + orientation.angles.x);
```
DonBruce64, line 1120: This also has the multiple != null checks as the other code, please convert to similar style.

DonBruce64 (changes requested):

PR #1682, opened 2023-07-11
1.18.2 port

gohkenytp:


DonBruce64 (changes requested):
None of this code complies.  There are import errors in multiple classes which, while some may be due to changes in the package names depending on your mappings, some are just in the wrong packages.  It is clear this "port" is just a copying of the 1.16.5 code and re-naming a bunch of functions, but not doing the whole porting.

image

gohkenytp:
Eesh. I feel bad for myself.

PR #1709, opened 2023-08-24
General Gradle Update and 1.20.1 Setup

leartpro:
The source code corresponds to that of 1.16.5 and is therefore partially faulty in 1-20.1. 
Furthermore, JEI cannot be found.


DonBruce64:
So you're saying you just copied the 1.16.5 source code, changed the numbers to 1.12.1. and did nothing else and therefore the code does not compile and a jar cannot be built?

PR #1716, opened 2023-09-12
Adds `gun_ammo_count_reloading`

boot2big:
Adds `gun_ammo_count_reloading` for checking the ammo count of whichever bullet is being loaded into the gun

Handy when used alongside VMs for different magazine models

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
@@ -1176,6 +1176,8 @@ public class PartGun extends APart {
                 return reloadTimeRemaining > 0 ? 1 : 0;
             case ("gun_ammo_count"):
                 return bulletsLeft;
+            case ("gun_ammo_count_reloading"):
+                return reloadingBullet.definition.bullet.quantity;
             case ("gun_ammo_percent"):
                 return bulletsLeft / definition.gun.capacity;
             case ("gun_active_muzzlegroup"):


LauraDarkez:
a very much appreciated feature, for guns that can take more than one count of ammunition, which is a good chunk of magazine fed guns, especially aks, ars, and some standardized pistols

boot2big:
Crap! That is supposed to be a separate PR. Eeeerrrr...

boot2big:
### Adds various VMs to PartEngine class
StartRPM VM
StallRPM VM
StarterPower VM
WearFactor VM
WinddownRate VM
RevlimitBounce VM
RevResistance VM
engine_rpm_idle
engine_rpm_start
engine_rpm_stall
engine_starter_power
engine_temp_ambient

boot2big:
And we'll all float on okay!

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
```diff
@@ -503,12 +520,15 @@ public void update() {
...
+                            } else if (!isActive) {
+                                stallEngine(Signal.INACTIVE);
+                                InterfaceManager.packetInterface.sendToAllClients(new PacketPartEngine(this, Signal.INACTIVE));
                             }
                         }
                     } else {
                         //Turn on engine if the magneto is on and we have fuel.
                         if (!world.isClient() && !vehicleOn.outOfHealth) {
-                            if (vehicleOn.isCreative || ConfigSystem.settings.general.fuelUsageFactor.value == 0 || vehicleOn.fuelTank.getFluidLevel() > 0) {
+                            if (vehicleOn.isCreative || ConfigSystem.settings.general.fuelUsageFactor.value == 0 || vehicleOn.fuelTank.getFluidLevel() > 0 || isActive == true) {
```
DonBruce64, line 531: isActive == true

Please don't check if boolean == true, that's asking IF true == true.  Every time you do that, a kitten cries.  Just do IF isActive

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
boot2big: Oops, it made sense in my head at the time... sorry. Will fix!

DonBruce64: So I checked this, and a question arises.  Will this code try and start vehicles in non-creative mode with 0 fuel?  Because looking at the logic, it looks like they'll start 
IF
they are creative OR they have FC of 0 OR they have fuel OR they are active.

So a vehicle without fuel, but still active, would attempt to start.  Mind you, it'd die the next tick for being outta fuel, but I suspect this would cause sound issues.  I think you want to start if the old conditions are present AND if the engine is active, so you'd want this.

isActive && (vehicleOn.isCreative || ConfigSystem.settings.general.fuelUsageFactor.value == 0 || vehicleOn.fuelTank.getFluidLevel() > 0)

boot2big: ... D'oh!

boot2big: Good catch, thank you for noticing that...

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
```diff
@@ -503,12 +520,15 @@ public void update() {
...
+                            } else if (!isActive) {
+                                stallEngine(Signal.INACTIVE);
+                                InterfaceManager.packetInterface.sendToAllClients(new PacketPartEngine(this, Signal.INACTIVE));
                             }
                         }
                     } else {
                         //Turn on engine if the magneto is on and we have fuel.
                         if (!world.isClient() && !vehicleOn.outOfHealth) {
-                            if (vehicleOn.isCreative || ConfigSystem.settings.general.fuelUsageFactor.value == 0 || vehicleOn.fuelTank.getFluidLevel() > 0) {
+                            if (isActive && vehicleOn.isCreative || ConfigSystem.settings.general.fuelUsageFactor.value == 0 || vehicleOn.fuelTank.getFluidLevel() > 0) {
```
DonBruce64, line 531: I think you fixed a syntax error here and un-fixed the code?  Now it will start if it's inactive, provided it has fuel.

boot2big: Arrrghh!! I am _not_ on a roll with this PR.

DonBruce64:
That outta did it!  Just make sure to wiki all over the place if you haven't already.

boot2big:
Naahh, I'mma kick it up a notch. BAM!

PR #1719, opened 2023-09-18
F you! *darkens your panel*

SnailPerson:
F you! *darkens your panel*

Darkened the base panel.png because it doesn't make sense to turn on a light that is already lit.
there's no light squared irl and especially not in minecraft!
It makes the use of the light switch in the panel more important.


boot2big:
Hmmm, guess it's a good thing my vehicles automatically turn on their panel lightup variable when you turn their engines on.

DonBruce64:
Are you sure this is the right thing to do?  It looks like you just removed all state-indicators from the panel when lights aren't on.

SnailPerson:
not too sure tbh... I like the idea of the nav-switch having more of an impact on the panel though...
but as of right now id say the panel is too dark, perhaps I can make the dark spots a bit brighter? 
To an extent where its still noticeable if the panel is on or off but visible enough to not be confusing when turned off.

DonBruce64:
Yeah.  I guess with me the thing is I'm thinking of cars.  When you're in a car, you don't notice much difference in daylight between having the vehicle lights on or off.  You instrument panel might get a bit darker if it's one of those fancy digital ones, but as to the analog ones?  There's actually almost no difference in the day.  At night though, it's very noticeable, since the panel is dark and therefore while the gauges are visible, they're not easy to see.  This is what I was going for with the lights, in that if you don't have them when it's dark you won't be able to see much of anything, but you can see things when it's daytime no matter what.

SnailPerson:
new dark variant:
panel



even newer, brighter variant:
panel

old variant:
panel

tried to find a proper middle ground

DonBruce64:
@boot2big thoughts about brightness?  Can you see the state of the buttons easily?

boot2big:
This thing did _not_ notify me. I'd have to test ingame, but at least here on the ol' site the brighter variant looks readable(... lookable?) enough to me!

SnailPerson:
So?

SnailPerson:
this should do the trick i think i hope

SnailPerson:
help

PR #1726, opened 2023-10-08
Mr.Mulles objectively superior custombench

SnailPerson:
Objectively superior custom bench.
I mean come on have you seen the old one?
wdwdawdad
Who ordered stapled meat cubes?

Have this instead!
awdawdd
aaddada
Much darker, solid, textured, sexy... those angles...
Thanks to chad succulent man for providing the wonderful model :D


DonBruce64:
A question here: can you place that bench down with the other benches?  I would think you'd want a bit brighter red for that one since it's more like the Craftsman toolboxes.  In fact, most toolboxes are brighter, so I would think the red could use a bit of that?

SnailPerson:
adawdad
current look

SnailPerson:
addwwadadwdaw
compared to other benches

SnailPerson:
QSqsSsqs

I could of course give it the palette of the fuel pump, but that one is too bright for my taste, the Bech seems to loose depth that way

DonBruce64:
That's a fair point.  Send it as-is then.

DonBruce64:
But, I see you changed the seatbench icon, was that because it got forget with the seatbench change before?

SnailPerson:
yep, forgor to do that int he last pr.
Texture with brighter colors is already in, so all you gotta do is press that merge button ;3

PR #1731, opened 2023-11-21
Fixed crash caused by referencing `gun_ammo_count_reloading` on a gun that's never been reloaded before

boot2big:
\*sigh\* Should've caught and fixed this _before_ 22.12.0 released, not after.

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
@@ -1177,7 +1177,9 @@ public class PartGun extends APart {
             case ("gun_ammo_count"):
                 return bulletsLeft;
             case ("gun_ammo_count_reloading"):
-                return reloadingBullet.definition.bullet.quantity;
+                return reloadTimeRemaining > 0 ? reloadingBullet.definition.bullet.quantity : 0; //EEEEVERYTHING YOU KNOW IS WRONG, BLACK IS WHITE UP IS DOWN AND INTS CAN'T 'null'!!! -cbc
+                //I would think that ideally this should've looked like the following, but apparently that's not actually possible, nor was it the solution anyways.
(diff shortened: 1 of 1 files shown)


mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -1177,7 +1177,9 @@ public double getRawVariableValue(String variable, float partialTicks) {
             case ("gun_ammo_count"):
                 return bulletsLeft;
             case ("gun_ammo_count_reloading"):
-                return reloadingBullet.definition.bullet.quantity;
+                return reloadTimeRemaining > 0 ? reloadingBullet.definition.bullet.quantity : 0; //EEEEVERYTHING YOU KNOW IS WRONG, BLACK IS WHITE UP IS DOWN AND INTS CAN'T 'null'!!! -cbc
+                //I would think that ideally this should've looked like the following, but apparently that's not actually possible, nor was it the solution anyways.
+                //return reloadingBullet.definition.bullet.quantity != null ? 1 : 0;
```
DonBruce64, line 1182: See, this is Java being slightly different than you'd expect.  The NullPointerError you see when you get a crash isn't because quantity is null, it's because reloadingBullet is null.  So what you should be checking for null is reloadingBullet.  If the bullet itself is null, then any time you try and check anything on the bullet, it will complain about null since the thing it's on doesn't exist.  It's like getting the fuel level of a fuel tank for a vehicle that doesn't exist.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
boot2big: Huh. Noted... will try and re-account for this!

PR #1732, opened 2023-11-23
Make "Welcome to MTS VERSION:" Info instead of Error

strubium:
Why? i do not know.


DonBruce64 (changes requested):
This change doesn't compile.

SnailPerson:
wonderful

PR #1736, opened 2023-12-08
Steering QOL improvement (configurable steering centering rate)

BlazeWolf1122:
This is a feature ive been wanting for as long as ive used MTS and ive finally been able to uhhh "do it"
I hope this isn't considered "janky" or "bad code", cause it does look a bit out of place when you look at the rest of the surrounding code


mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
```diff
@@ -64,7 +64,7 @@ public class EntityVehicleF_Physics extends AEntityVehicleE_Powered {
     public double rudderTrim;
     public static final double MAX_RUDDER_ANGLE = 45;
     public static final double MAX_RUDDER_TRIM = 10;
-    public static final double RUDDER_DAMPEN_RATE = 2.0;
+    public static final double RUDDER_DAMPEN_RATE = ConfigSystem.client.controlSettings.steeringReturnRate.value;
```
DonBruce64, line 67: This won't do what you want it to do.  This only checks the config once when you boot, and won't check if you change the config mid-game.  What you want to do is not use this variable on client-side control inputs, but leave it for server-side inputs.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/systems/ControlSystem.java
```diff
@@ -489,7 +489,11 @@ private static void controlGroundVehicle(EntityVehicleF_Physics powered, boolean
 
         //Check steering.  Don't check while on a road, since we auto-drive on those.
         if (!powered.lockedOnRoad) {
-            controlControlSurface(powered, ControlsJoystick.CAR_TURN, ControlsKeyboard.CAR_TURN_R, ControlsKeyboard.CAR_TURN_L, ConfigSystem.client.controlSettings.steeringControlRate.value, EntityVehicleF_Physics.MAX_RUDDER_ANGLE, EntityVehicleF_Physics.RUDDER_INPUT_VARIABLE, powered.rudderInput, EntityVehicleF_Physics.RUDDER_DAMPEN_RATE);
+            if (!ControlsKeyboard.CAR_TURN_L.isPressed() && !ControlsKeyboard.CAR_TURN_L.wasPressedLastCall && !ControlsKeyboard.CAR_TURN_R.isPressed() && !ControlsKeyboard.CAR_TURN_R.wasPressedLastCall) {
```
DonBruce64, line 492: Very close sir.  I see you have both paramters in the config.  But you don't reference the custom dampen rate here, so I think you want this in this code section?  See how the controlSettings.steeringControlRate.value is used for the increment rate?  You need to feed in your new config rather than the RUDDER_DAMPEN_RATE to make this work.  No if-pressed-checks required.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/systems/ControlSystem.java
BlazeWolf1122: Thanks for letting me know, will try to get to it in a few days

BlazeWolf1122:
hello this took far too long but i finally used your advice and updated the PR
this has been haunting me in my dreams

BlazeWolf1122:
also im pleasantly surprised that the if-then thing could've simply just been one line
well, i guess you don't know until you know

PR #1740, opened 2023-12-11
Gravity Factor and Climb Speed for Vehicle JSON

Naxgeneral:
This is feature where you can edit the Gravity Factor and Climb Rate inside Vehicle Motorized section inside json. 
as the default fall speed for 0.35 was too low and i want it to be edit able in each vehicle
how it work? : it will check does "gravityFactor" != 0 in vehicle json  if it == 0 it will take value from Config file else it will take value from "gravityFactor"

and for climbSpeed (climbRate) i feel like default wheel climb speed also too slow for 0.35 and manytime i ended up stuck in slope so i make it to be edit able in json motorized section 
it work the same as gravity but for climb speed.


DonBruce64:
Looking at this, it makes me wonder: should I just ignore speedFactor for climbFactor?  make vehicles climb at 1.0 speed?  And for cars (not planes) should they always fall at 1.0 speed?  Seems better to do it for all vehicles so packs don't all have to update VS having is configurable.

What factors do you see that work well for you when you are playing?  2.5/3.0 I would think?

DonBruce64 (changes requested):
Looking at this, it makes me wonder: should I just ignore speedFactor for climbFactor? make vehicles climb at 1.0 speed? And for cars (not planes) should they always fall at 1.0 speed? Seems better to do it for all vehicles so packs don't all have to update VS having is configurable.

What factors do you see that work well for you when you are playing? 2.5/3.0 I would think?

Naxgeneral:
For the Climb factor i added it for fixing vehicle dig into the ground when climbing slope at high speed, also i make if else to check for both of them presenting in JSON(if those value present just use them but if not it will use default value) so other PA dont have to rewrite their own json

DonBruce64:
Same comment though.  What factors do you use for your vehicles normally?  I would think something close to 2.8?

Naxgeneral:
> ignore speed factor for climb factor

A: Kinda works 

> should they always fall at 1.0 speed? Seems better to do it for all vehicles so packs don't all have to update VS having is configurable.

A: Excuse me for not understand what VS is but in my code i have if else to check are the factor value being set. If not it will just load value from main confi file but if it being set it will take that value in json  no need to add those climbFactor or gravityFactor to your json if you need the value from config

But meanwhile a lot of if else can cause performance issue so im not sure  and making all vehicle exclude plane fall at 1:1 is sound nice to me bt for other pack i dont know  so that why i want it to be definable in json

> What factors do you see that work well for you when you are playing? 2.5/3.0 I would think?

A: for climb i usually play around 1.0 to 2.0
For gravity im around 2 to 3

Naxgeneral:
> Same comment though. What factors do you use for your vehicles normally? I would think something close to 2.8?

Yes around that

DonBruce64:
>A: Excuse me for not understand what VS is but in my code i have if else to check are the factor value being set. If not it will just load value from main confi file but if it being set it will take that value in json no need to add those climbFactor or gravityFactor to your json if you need the value from config

So, what I was saying is this:

Normally, vehicles in IV will move at speedFactor speed.  So if you set it to 0.35, they move at 0.35 speed.  This also is for climbing up blocks.  They move up at 0.35 speed.  So if you make them climb normally at 1.0 speed, then they will climb a lot faster.

Gravity is the same.  Cars are "floaty" cause they fall at 0.35 speed.  Same for planes which fall slow, but they should fall slower because they also fly slower and climb slower.  So I don't want to mess with plane gravity, only car gravity.  So what I am thinking is this:

1) Make climb speed not care about speedFactor.  doing that will make it essentially have a value of 2.8 with your code because 1/0.35 = 2.8.

2) Make gravity be 1.0 for vehicles (factor of 2.8 with speedFactor 0.35), but only if the vehicle is not isAircraft or isBlimp.

Naxgeneral:
> Normally, vehicles in IV will move at speedFactor speed. So if you set it to 0.35, they move at 0.35 speed. This also is for climbing up blocks. They move up at 0.35 speed. So if you make them climb normally at 1.0 speed, then they will climb a lot faster.

So the thing is i make vehicle go 1:1 speed at 0.35 thats why i got floaty and digging car issue

Also your ideas are also works

DonBruce64:
> So the thing is i make vehicle go 1:1 speed at 0.35 thats why i got floaty and digging car issue

That makes sense then.  The proposed change will work then.  Not sure if you want to try your hand at a PR for this or spend some coin for this as a feature request.  Either way, do please open another PR ticket or issue ticket and we'll make it happen.  Easy to do.

Naxgeneral:
Actually for this PR i already have working code and tested to be work fine but if you wanted that idea i can do other try on PR

DonBruce64:
I would rather the issue be fixed in the code without a per-pack config since then all packs get it fixed.  Doing this per-pack means that other, older, packs won't get the benefit of faster climbing/gravity unless they update each vehicle.  And I'd rather just have better code than have it be something each pack fixes individually.

Naxgeneral:
Ohh i understand now why you dont what that, so i will do PR with the fixed one. but what if some people with 0.35 default speed dont like it? how about i add fixed code and then take back that if else if you want to tweak it some more?

DonBruce64:
I don't think folks with 0.35 will complain.  I personally feel that cars should probably go up slopes quicker anyways after playing with them a bit, and certianly down via gravity.  Driving off a building seems... wrong... with IV right now.

PR #1741, opened 2023-12-11
Propeller Pitch Speed Config

Naxgeneral:
Add ability to edit Dynamic Speed Propeller pitch changing speed in part JSON file as "pitchSpeed"

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartPropeller.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartPropeller.java
@@ -40,6 +40,10 @@ public class PartPropeller extends APart {
      * In meters per second.
      **/
     public double desiredLinearVelocity;
+    /**
+     * JSON Value.
+     **/
+    public int pitchSpeed = definition.propeller.pitchSpeed;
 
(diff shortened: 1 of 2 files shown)


mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartPropeller.java
```diff
@@ -131,9 +135,21 @@ public void update() {
                     }
                 } else if (definition.propeller.isDynamicPitch) {
                     if (decreasePitch || (vehicleOn.reverseThrust && currentPitch > -MIN_DYNAMIC_PITCH)) {
-                        --currentPitch;
+                        if(pitchSpeed != 0){
```
DonBruce64, line 138: Rather than check for this, add a line in LegacyCompatSystem to make this 1 if the pitchSpeed is 0 and the propeller is dynamic pitch.

Also, when you do that, just check if pitchSpeed is not 0 and do this logic.  No need to check if it's dynamic pitch if you have a pitchSpeed instead.

DonBruce64 (changes requested):

DonBruce64:
Closed due to feature being added.

PR #1742, opened 2023-12-14
Allowed constantValues to be defined on a part slot

DrprofLuigi:
When a constantValues list is defined on a part slot, it is appended to that part's internal constantValues.


DonBruce64:
Looks good to me.  Just make sure to update the wiki.

PR #1745, opened 2023-12-16
Fix the build script (finally)

Elephant1214:


Elephant1214:
I just remembered (after reading through this) that I moved the version modification to the subprojects, I'll add that tomorrow.

PR #1770, opened 2024-02-04
Add Anim variable for flaps

conman180:
Add Booleans for flap animaions. that way it can be determined if flaps are increasing in angle (extending) or decreasing in angle (retracting).


conman180:
Will add wiki entries once merged

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
```diff
@@ -861,6 +861,10 @@ public double getRawVariableValue(String variable, float partialTicks) {
                 return flapCurrentAngle;
             case ("flaps_moving"):
                 return flapCurrentAngle != flapDesiredAngle ? 1 : 0;
+            case ("flaps_increasing"):
+                return flapCurrentAngle != flapDesiredAngle && flapCurrentAngle < flapDesiredAngle ? 1 : 0;
```
DonBruce64, line 865: You don't need the not-equal check here, since it is redundant if you check if it is less than.  It cannot be less than, and equal to, so the first check is not needed.

DonBruce64 (changes requested):

PR #1778, opened 2024-03-06
steeringForceFactor + steeringForceIgnoresSpeed

boot2big:
### This PR does four significant things:
- downForce has been renamed to steeringForceFactor, otherwise retaining its same functionality\* _(\*See below)_
- steeringForceIgnoresSpeed, switches handling behavior to ignore the vehicle's current velocity. 
- When above boolean is true, ~~downForce~~ steeringForceFactor directly controls the vehicle's turningForce instead of it being influenced by speed and steeringForceFactor.
- slip_understeer, current steering angle - turningForce (if the vehicle cannot turn at all this returns 1.0, otherwise everything between 0.0)

### But why?
Some pack authors may find the current behavior of downForce/overSteer/underSteer to offer less control over their vehicle's handling than wished. This is especially noticeable because at low speeds, downForce has little to no effect on vehicle handling. However this pull request seeks to alleviate that by retaining this behavior by default whilst offering an alternative.
As ~~downForce~~ steeringForceFactor can be variableModified, this results in more customizable vehicle behavior through the use of clever VMs.

### Other stuff
downForce has technically been deprecated, however it is immediately LC'ed to steeringForceFactor. As there is no change in functionality when steeringForceIgnoresSpeed is false, this has no effect on any packs to my knowledge.
Demonstrational Video

https://github.com/DonBruce64/MinecraftTransportSimulator/assets/53581325/dfe40d9d-b263-4e3a-b672-5b83441816fc

\*sigh\* Point, not ponit.

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
@@ -354,7 +354,7 @@ abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding {
                     vectorDelta = 180 + vectorDelta;
                 }
             }
-            if (this.towedByConnection == null) {
+            /*if (this.towedByConnection == null) {
                 double overSteerForce = Math.max(velocity / 4, 1);
                 if (definition.motorized.overSteerAccel != 0) {
                     weightTransfer += ((motion.dotProduct(motion, false) - prevMotion.dotProduct(prevMotion, false)) * weightTransfer) * currentOverSteer;
@@ -367,7 +367,7 @@ abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding {
                     weightTransfer = currentOverSteer;
                 }
                 rotation.angles.y += crossProduct.y * weightTransfer + (Math.abs(crossProduct.y) * -currentUnderSteer * turningForce) * overSteerForce;
-            }
+            }*/
 
             //If we are offset, adjust our angle.
             if (Math.abs(vectorDelta) > 0.001) {
@@ -519,14 +519,14 @@ abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding {
                 //This is opposite of the torque-based forces for control surfaces.
                 double turningForce = steeringAngle / turningDistance;
                 //Decrease force by the speed of the vehicle.  If we are going fast, we can't turn as quickly.
-                if (groundVelocity > 0.35D) {
-                    turningForce *= Math.pow(0.3F, (groundVelocity * (1 - currentDownForce) - 0.35D));
-                }
+                /*if (groundVelocity > 0.35D) {
+                //    turningForce *= Math.pow(0.3F, (groundVelocity * (1 - currentDownForce) - 0.35D));
+                }*/
                 //Calculate the force the steering produces.  Start with adjusting the steering factor by the ground velocity.
                 //This is because the faster we go the quicker we need to turn to keep pace with the vehicle's movement.
                 //We need to take speed-factor into account here, as that will make us move different lengths per tick.
                 //Finally, we need to reduce this by a constant to get "proper" force..
-                return turningForce * groundVelocity * (speedFactor / 0.35D) / 2D;
+                return turningForce * (groundVelocity * (currentOverSteer + 1)) * (speedFactor / 0.35D) / 2D;
             }
         }
         return 0;
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
@@ -69,7 +69,9 @@ abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding {
 
     //Properties
     @ModifiedValue
-    public float currentDownForce;
+    public float currentSteeringForceIgnoration;
+    @ModifiedValue
+    public float currentSteeringForceFactor;
     @ModifiedValue
     public float currentBrakingFactor;
     @ModifiedValue
@@ -354,7 +356,7 @@ abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding {
                     vectorDelta = 180 + vectorDelta;
                 }
             }
-            /*if (this.towedByConnection == null) {
+            if (this.towedByConnection == null) {
                 double overSteerForce = Math.max(velocity / 4, 1);
                 if (definition.motorized.overSteerAccel != 0) {
                     weightTransfer += ((motion.dotProduct(motion, false) - prevMotion.dotProduct(prevMotion, false)) * weightTransfer) * currentOverSteer;
@@ -367,7 +369,7 @@ abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding {
                     weightTransfer = currentOverSteer;
                 }
                 rotation.angles.y += crossProduct.y * weightTransfer + (Math.abs(crossProduct.y) * -currentUnderSteer * turningForce) * overSteerForce;
-            }*/
+            }
 
             //If we are offset, adjust our angle.
             if (Math.abs(vectorDelta) > 0.001) {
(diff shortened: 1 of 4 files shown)


mccore/src/main/java/minecrafttransportsimulator/packloading/LegacyCompatSystem.java
```diff
@@ -361,9 +361,9 @@ private static void performVehicleLegacyCompats(JSONVehicle definition) {
             definition.motorized.hasCruiseControl = false;
         }
 
-        //If downforce is greater than 1, set it to 0.5 to prevent spinny vehicles.
-        if (definition.motorized.downForce > 1) {
-            definition.motorized.downForce = 0.5F;
+        //Change downForce's name to steeringForceFactor
```
DonBruce64, line 364: You'll want to also set downForce to 0 here to prevent it and steeringForceFactor from showing up in the devmode outputted JSON.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
```diff
@@ -877,6 +881,9 @@ public double getRawVariableValue(String variable, float partialTicks) {
                 return (rotation.angles.x) / 0.15F * 25F;
             case ("slip"):
                 return 75 * sideVector.dotProduct(normalizedVelocityVector, true);
+            case ("slip_understeer"):
```
DonBruce64, line 884: This section is a control-flow error.  Though not an obvious one.  If you have a case statement without curly bois {} after it, it only executes the first block of code, then goes on to do other stuff.  So you're just calculating turning force here and returning 0 later on down the line.  Either add curlys, or make it a one-liner like the rest of the entries here.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
```diff
@@ -69,7 +69,9 @@ abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding {
 
     //Properties
     @ModifiedValue
-    public float currentDownForce;
+    public float currentSteeringForceIgnoration;
```
DonBruce64, line 72: Request name of currentSteeringForceIgnored

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
```diff
@@ -422,7 +424,7 @@ private float getBrakingForce() {
      * device calculations should be applied due to said devices being in
      * contact with the ground.
      */
-    private float getSkiddingForce() {
+    protected float getSkiddingForce() {
```
DonBruce64, line 427: This should still be private VS protected?

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
```diff
@@ -439,7 +441,7 @@ private float getSkiddingForce() {
      * Sign of returned value indicates which direction entity should yaw.
      * A 0 value indicates no yaw change.
      */
-    private double getTurningForce() {
+    protected double getTurningForce() {
```
DonBruce64, line 444: This should still be private VS protected?

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
```diff
@@ -519,14 +521,16 @@ private double getTurningForce() {
...
+                    turningForce *= Math.pow(0.3F, (groundVelocity * (1 - currentSteeringForceFactor) - 0.35D));
+                } else if (currentSteeringForceIgnoration != 0) {
+                    turningForce *= currentSteeringForceFactor;
                 }
                 //Calculate the force the steering produces.  Start with adjusting the steering factor by the ground velocity.
                 //This is because the faster we go the quicker we need to turn to keep pace with the vehicle's movement.
                 //We need to take speed-factor into account here, as that will make us move different lengths per tick.
                 //Finally, we need to reduce this by a constant to get "proper" force..
-                return turningForce * groundVelocity * (speedFactor / 0.35D) / 2D;
+                return turningForce * (groundVelocity * (currentOverSteer + 1)) * (speedFactor / 0.35D) / 2D;
```
DonBruce64, line 533: I see you added currentOverSteer here.  Was this intentional?  I know we already use this value somewhere else given that you didn't add it to the class file, so isn't it doing what it's supposed to somewhere else?

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/packloading/LegacyCompatSystem.java
boot2big: Ah? I thought this wasn't necessary at first, but now that you point it out that makes sense if I have to define it as deprecated. Thank you!

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
boot2big: I.. thought it worked when I tested it? It still returned the result from what I could see, but I'll tweak this as advised. :+1:

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
boot2big: `steeringForceIgnoresSpeed`
Was unsure if `currentSteeringForceIgnoresSpeed` would fit better. EZ change if so

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
boot2big: That's a mistake on my part in figuring out how to return understeer, my apologies. Will be privated.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
boot2big: _This_ I did intentionally open because that is what gets used in `slip_understeer`. If getTurningForce() is supposed to be private, should I simply make a public variable that returns getTurningForce()?

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
boot2big: Eeehm, no it is not..? That's another mistake on my part, I haven't the foggiest how that got added. Might've happened while I was crossreferencing a fellow PA's PRs, as the precursor to this request was me trying to rip the code apart to replace it entirely.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
DonBruce64: You're right, it should be prefixed with `current` as it's a ModifiedValue.  So `currentSteeringForceIgnoresSpeed`  it should be.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
DonBruce64: Ah, I missed that, makes sense.  What I'd do here is make the `turningForce` variable located here https://github.com/DonBruce64/MinecraftTransportSimulator/blob/a61731231e7a2ac327db4fb69bcbf631b44db042/mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java#L330C16-L330C29

a variable in the class file up top that's `protected` and reference that in your `slip_understeer`.  that way you aren't re-calculating the turning force each time you reference the variable, you're using the last-calculated value that's known correct.  Saves on processing time, and lets this method remain private as a helper method.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
DonBruce64: It might be me being mistaken, but I coulda swore that's how case statements work since IF statements without curlys do the same thing.  So I keep them the same.  Either way, it's for sure fixed now!

DonBruce64:
Merged.  Go forth and wiki this!

PR #1780, opened 2024-03-11
fix typo for second time. didnt merge right i guess

conman180:


conman180:
i have no clue why it didnt merge properly before

PR #1789, opened 2024-04-15
Add text object definition "variableOffset"

conman180:
Can be useful in text objects where rounding causes the floating point to increment before it should. example video on discord.


DonBruce64:
I'm surprised I didn't put such a thing in before.  Having an offset it something you can have on other variables.  Make sure to update the Wiki!

PR #1798, opened 2024-06-02
Mr.Mulles Acoustic Benches - Episode 1

SnailPerson:
Added handmade sounds to most of the current cored benches, utilizing the previously added "player_interacting" variable.
Now fancy sounds will play when opening a benches GUI, while having the GUI open, when closing a benches GUI and when crafting an item of course!

Current Acoustic support:
- ✅ custom bench
- ✅ decor bench
- ✅ gun bench
- ✅ instrument bench
- ✅ item bench
- ✅ seat bench

Coming soon:
- propeller bench
- engine bench
- vehicle bench
- wheel bench


dldev32:
> @SnailPerson So, a question, how much additional file size do these files take up? I want to make sure the core mod stays under 8MB since that's the Discord size limit.

But the file limit in discord is 25 MB, or am I missing something?

SnailPerson:
@DonBruce64 as of now, the additional content roughly adds up to around 0.75 MB

PR #1799, opened 2024-06-03
add guidance delay

conman180:


conman180:
ok no idea why all those other merged commits are still there

PR #1801, opened 2024-06-03
Add slip Degrees animation variable

conman180:
i used your formula for finding "trackAngle" aka angle of attack but applied it to slip to give a degree value between the 2 vectors used in the dot product function.

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
@@ -893,7 +893,7 @@ public class EntityVehicleF_Physics extends AEntityVehicleE_Powered {
             case ("slip"):
                 return 75 * sideVector.dotProduct(normalizedVelocityVector, true);
             case ("slip_degrees"):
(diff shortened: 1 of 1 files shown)


mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
```diff
@@ -892,6 +892,8 @@ public double getRawVariableValue(String variable, float partialTicks) {
                 return (rotation.angles.x) / 0.15F * 25F;
             case ("slip"):
                 return 75 * sideVector.dotProduct(normalizedVelocityVector, true);
+            case ("slip_degrees"):
+                return -Math.toDegrees(Math.asin(sideVector.dotProduct(normalizedVelocityVector, true)));
```
DonBruce64, line 896: I don't think you need to pass-in true as an argument here?  IIRC, that's to normalize the inputs, which since they're already normalized, just does extra processing that's not required.

DonBruce64 (changes requested):

DonBruce64:
Please update the wiki with this information.

PR #1804, opened 2024-06-08
Add changes to gravity

conman180:
For cars, Normalizes gravity to be 100% speed no matter the speed factor.
For anything else, allows gravity factor to be defined/modified thru VMs as motorized section definition.

Will make it so gravity factor applies on top of normalized gravity for cars upon request.


mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
```diff
@@ -550,12 +556,19 @@ protected void getForcesAndMotions() {
 
             //Finally, get gravity.
             gravitationalForce = currentBallastVolume == 0 ? currentMass * (9.8 / 400) : 0;
+            /*if (definition.motorized.isAircraft) {
```
DonBruce64, line 559: I spy dead code...

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
```diff
@@ -550,12 +556,19 @@ protected void getForcesAndMotions() {
...
+            } else {
+                gravitationalForce = currentBallastVolume == 0 ? currentMass * (9.8 / 400) / speedFactor : 0;
+            }*/
             if (currentWaterBallastFactor != 0 && world.isBlockLiquid(position)) {
                 gravitationalForce -= gravitationalForce * currentWaterBallastFactor;
                 elevatorTorque = -orientation.angles.x * 2;
                 aileronTorque = -orientation.angles.z * 2;
             }
-            if (!definition.motorized.isAircraft) {
+            if (definition.motorized.gravityFactor != 0) {
```
DonBruce64, line 569: I think you should be checking the currentGravityFactor here, not the JSON one.  If someone has 0 in the JSON, but is using VMs instead to modify it, then it won't get applied here even if it's a non-zero value.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
```diff
@@ -550,12 +556,19 @@ protected void getForcesAndMotions() {
...
+                gravitationalForce = currentBallastVolume == 0 ? currentMass * (9.8 / 400) : 0;
+            } else {
+                gravitationalForce = currentBallastVolume == 0 ? currentMass * (9.8 / 400) / speedFactor : 0;
+            }*/
             if (currentWaterBallastFactor != 0 && world.isBlockLiquid(position)) {
                 gravitationalForce -= gravitationalForce * currentWaterBallastFactor;
                 elevatorTorque = -orientation.angles.x * 2;
                 aileronTorque = -orientation.angles.z * 2;
             }
-            if (!definition.motorized.isAircraft) {
```
DonBruce64, line 558: You removed the aircraft check in your updated code, that means that aircraft now are subject to the gravity config.  That was left out specifically because aircraft are meant to be floatier than cars.  Cars should have more gravity, but they don't because they factor the Y value by the speedFactor, which makes them go down slower.  that's ideal for planes where you need to shorten takeoff roll, but not cars where you want them to always fall down fast, just not move fast.  You'd want to check the D_Moving class where the factor is applied, and have it ignore the factor in the y-dir for cars, but not aircraft/blimps.

DonBruce64 (changes requested):

DonBruce64:
Actually, does your change supercede Nax's #1792 ?

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
conman180: the reason i had it like this is so that if left undefined in json, it resorts to global of 1 or whatever mts config has set. i dont think having zero gravity vehicles is a good idea

conman180:
> Actually, does your change supercede Nax's #1792 ?

only partially. the gravity stuff yes, wheel stuff no

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
conman180: i intended to remove aircraft check in code to have it possible to modify how it acts on a plane. making it customisable vehicle to vehicle. an example use would be a totaled/shotdown plane not wanting to be floaty as it cant fly no more. so PA can make it come down quicker if they so choose.
hence why prev stated, i left it so a 0 or undefined resorts back to original behavior.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
conman180: yeah yeah ill get to this

conman180:
Temporarily closing this for now to update my fork

PR #1807, opened 2024-06-23
Patch 1

BlazeWolf1122:
big thanks to conman for helping me fix up my jank ass code cause he knows java and i dont
Adjustable thrust vectoring value for engine parts, as thrust changes to mts physics system with thrust vectoring has proven to be a headache for me and other PAs, especially those from GAP. There are no LCs for defaulting the value to 1.0, as a majority of people who ive asked in mts were in favor of defaulting the value to 0 when thrust vectoring is not found. Also, this jet thrust vectoring is VMable Please let me know in MTS if theres any issues or code oversights.

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
@@ -111,6 +111,7 @@ public class PartEngine extends APart {
     private double prevDriveshaftRotation;
     private double currentJetPowerFactor;
     private double currentBypassRatio;
+    private double currentThrustVectorMultiplier;
     private final List<PartGroundDevice> linkedWheels = new ArrayList<>();
     private final List<PartGroundDevice> drivenWheels = new ArrayList<>();
     private final List<PartPropeller> linkedPropellers = new ArrayList<>();
@@ -784,7 +785,7 @@ public class PartEngine extends APart {
         currentWinddownRate = definition.engine.engineWinddownRate;
         currentJetPowerFactor = definition.engine.jetPowerFactor;
         currentBypassRatio = definition.engine.bypassRatio;
-
+        currentThrustVectorMultiplier = definition.engine.thrustVectorMultiplier;
 
         //Adjust current variables to modifiers, if any exist.
         if (definition.variableModifiers != null) {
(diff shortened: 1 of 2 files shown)


mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
```diff
@@ -1305,7 +1311,7 @@ public double addToForceOutput(Point3D force, Point3D torque) {
             engineForce.set(engineAxisVector).scale(thrust);
             force.add(engineForce);
             engineForce.reOrigin(vehicleOn.orientation);
-            torque.add(localOffset.crossProduct(engineForce));
+            torque.add(localOffset.crossProduct(engineForce).scale(currentThrustVectorMultiplier));
```
DonBruce64, line 1314: You mentioned this `There are no LCs for defaulting the value to 1.0, as a majority of people who ive asked in mts were in favor of defaulting the value to 0 when thrust vectoring is not found.`

This means that torque will never be added by default, where it was before, since all engine torque will be scaled by 0, and multiplying anything by 0 makes it 0.  Are folks SURE they want to remove thrust-torquing by default?  This will result in jets with an engine out behaving differently than propeller craft with engines out, since propellers will apply torque but jets won't.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
conman180: you wouldnt believe the storm of complaits about GAP right now due to engines being located away from model origin points. even though they shouldnt vector. the amount of torque being applied is huge in some cases leading to planes doing wheelies or stoppies just by throttling up and stationary

Deepseasaltyfish: well i think the problem may also due to mts doesnt  take into account the moment generated by lift and gravity, the list and gravity force are lift force are apply at the same position, so if the engine are lower or higher than z0, the thrust vectoring will make the plane pull its nose up or down even at very speed, maybe we need a CG system?
4ead5f0f10cdc596

PR #1819, opened 2024-07-18
Add furnace variables for interactable parts

boot2big:
interactable_furnace_fuel = Amount of fuel ticks remaining interactable_furnace_remaining = Ticks left until item is finished smelting/cooking


DonBruce64:
Surprised this wasn't in there already.  Added.

DonBruce64:
Wiki please when you get a moment!

boot2big:
I was surprised as well, but they were finally needed for a particular feature so that's why I went and added them.

Wiki'd

PR #1832, opened 2024-08-02
Feat: Make vehicles explode when hitting walls without breaking the walls

ktokto313:


DonBruce64:
Obsoleted via 331a81d1236e8b40baa0656788b82cf041cf4823

PR #1836, opened 2024-08-11
Retool to Essential Loom

Elephant1214:
Retool to Essential Loom, supporting up to Fabric 1.21.1 and Forge 1.21, and now requiring Java 21.

# Things to note
- A Java 21 JDK is REQUIRED, not optional
- 1.12.2 runs MUST USE JAVA 8 or they won't work
- Tasks on subprojects must be run through the main project using `gradlew :subprojectName:task`

# Problems
- Resources are not shaded into 1.12.2 and 1.16.5 projects when run in an IDE for some reason
- No way to automatically update version numbers (yet)
- Jar names are probably not to Don's liking, but this can be fixed easily
- Access Transformer AND Access Widener required for modern Forge, but apparently Essential Loom can convert these itself on the fly (?)

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/rendering/RenderableVertices.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/rendering/RenderableVertices.java
@@ -301,16 +301,24 @@ public class RenderableVertices {
     }
 
     /**
+     * Adds a line to the {@link #vertices} of this object using Point3D objects.
+     * If the last line is added, this function will automatically handle the batch ending.
+     */
+    public void addLine(Point3D start, Point3D end) {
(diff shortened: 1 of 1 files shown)


slava110:
Note that for 1.12.2 forge version will be downgraded from `14.23.5.2860` to `14.23.5.2840`. Though, I don't think there's a lot of important changes for dev environment between these versions

Elephant1214:
> Note that for 1.12.2 forge version will be downgraded from `14.23.5.2860` to `14.23.5.2840`. Though, I don't think there's a lot of important changes for dev environment between these versions

No difference for MTS. All I've found is that Mixin has a few problems with mappings, but using UniMix fixes them.

PR #1845, opened 2024-09-01
Add airDensityDeclineRate to config system

Deepseasaltyfish:
By consulting the data I think the realistic air density drops by about half every 6.7km,
but old file uses 0.5km so planes can never fly so high like in reality.
after improving climb porformance we can fly higher, having more fun with shader or something else


DonBruce64:
Two things:
1) the reason I check for max world height is to account for the difference in normal and cubicChunks worlds automatically.  Please keep this code in.

2) Why are you making this a config that most users won't even touch?  Why not just make it global?

Deepseasaltyfish:
for 1: cubicChunks players can set airDensityDeclineRate larger to allow them flying higher, 
and it should be easier for them to cauculate the number they need, the climb hight is usually between 1*airDensityDeclineRate and 2*airDensityDeclineRate.

for 2: make it global means just make it a constant? 
but for some reason players may need to fly higher or lower, they need different climb performance so i think make it configable is better

Deepseasaltyfish:
automatically adapting world height may make them confuse about how to calculate this value, 
also a multiple of world height cant determine how high player wants to fly, just leave sapce for them to config. 
and i will explain more clear in the config comment sentence

Deepseasaltyfish:
> Two things:
> 
> 1. the reason I check for max world height is to account for the difference in normal and cubicChunks worlds automatically.  Please keep this code in.
> 2. Why are you making this a config that most users won't even touch?  Why not just make it global?

also for future 1.17+ versions whose height limit is much higher, i believe most players prefer 1:1 climb height, whether keep getWorldHeight in or not depends on you lol

conman180:
if i may interject: i would suggest scaling the density decay rate by the current speed factor set in mtsConfig. Allowing for a proer decay based on the speed being scaled down. since speed facotr defaults to 0.35. you should only be able to fly 0.35* the height it really would if you get where im going with this. i do the same thing when configuring missile speeds

Deepseasaltyfish:
> if i may interject: i would suggest scaling the density decay rate by the current speed factor set in mtsConfig. Allowing for a proer decay based on the speed being scaled down. since speed factor defaults to 0.35. you should only be able to fly 0.35* the height it really would if you get where im going with this. i do the same thing when configuring missile speeds

emm I think that's not intuitive enough either, and the speedfactor was only supposed to be a compromise on computer performance, I don't see the need for it to affect flight altitude

DonBruce64:
> > if i may interject: i would suggest scaling the density decay rate by the current speed factor set in mtsConfig. Allowing for a proer decay based on the speed being scaled down. since speed factor defaults to 0.35. you should only be able to fly 0.35* the height it really would if you get where im going with this. i do the same thing when configuring missile speeds
> 
> emm I think that's not intuitive enough either, and the speedfactor was only supposed to be a compromise on computer performance, I don't see the need for it to affect flight altitude

Agreed.  speed factor is indeed only for computer performance, and is meant to laterally slow down planes and cars to not over-speed in the world.  Actual max altitude should remain un-changed.

Deepseasaltyfish:
> > also for future 1.17+ versions whose height limit is much higher, i believe most players still prefer 1:1 climb height, whether keep getWorldHeight in or not depends on you lol
> 
> I would like to keep the world height in there. If the world is higher in 1.17+, that bit of code will handle it. I'd like for planes to "cap out" at specific altitudes a little bit above the world. The idea being that jets will fly higher than prop craft. And some prop craft won't fly barely above the clouds due to their weaker engines.
> 
> If someone wants to adjust the config to make planes fly super high, then they can, and that's what the config file is for.
> 
> That being said, you have me curious, what's the height of the plane in the photo, what's the world max height, and what's the height of the clouds? Is that a mod that moves clouds super high up?

the plane can fly higher than 10km with value 6700, wold height is just 256 but clouds of shader can be much higher, the cloud height is between 1km and 5km(i dont remember the accurent number), also with a mod fixed optfine bug the plane wont disalpear higher than 256

DonBruce64:
Ah, I wasn't aware that that's where clouds were.  If that is the case, how about this: we change the function to a simple logarithmic one like how it is in real-life, and just make the config be where air density is 0?  Far simpler to do and understand and code mathamatically.  Linear function would be like this:

https://scied.ucar.edu/sites/default/files/interactives/atmosphere-altitude/surface-pressure-altitude-earth.jpg


As to why I did that, the thought process was that if you had a heavy plane loaded up, you'd struggle to get altitude since you'd not have the lifting potential required due to the low air density.  Plus, planes fly better faster up in the higher, less-dense, air due to lower wing forces.  But honestly, it's probably easier from the standpoint of end-users to just set the max flight height super high and leave it at that, rather than trying to give nuances to the physics.

@conman180  you're a pilot, thoughts?

Deepseasaltyfish:
> Ah, I wasn't aware that that's where clouds were. If that is the case, how about this: we change the function to a simple logarithmic one like how it is in real-life, and just make the config be where air density is 0? Far simpler to do and understand and code mathamatically. Linear function would be like this:
> 
> https://scied.ucar.edu/sites/default/files/interactives/atmosphere-altitude/surface-pressure-altitude-earth.jpg
> 
> As to why I did that, the thought process was that if you had a heavy plane loaded up, you'd struggle to get altitude since you'd not have the lifting potential required due to the low air density. Plus, planes fly better faster up in the higher, less-dense, air due to lower wing forces. But honestly, it's probably easier from the standpoint of end-users to just set the max flight height super high and leave it at that, rather than trying to give nuances to the physics.
> 
> @conman180 you're a pilot, thoughts?

this sounds better, and i find this: https://www.grc.nasa.gov/WWW/K-12/airplane/atmosmet.html
300f8236cea2c0e4
how about making a standard model and multiply a scale factor? also i think we can also do the same thing to temperature

PR #1875, opened 2024-12-27
Add "beacon_connected"

conman180:
animation variable to determine if a beacon is connected


conman180:
i have no idea why the older commits still show up even tho they were merged a while back

PR #1878, opened 2024-12-31
Seperate elevator control from ballast

boot2big:
Still keeps compatibility with blimps by defaulting to elevatorAngle when unmodified. However, the ballast difference can now be modified and overrode by VMs if needed

Allows for vehicles that can float, but also pitch independently of each other


DonBruce64:
Throw it on the wiki, tis merged!

PR #1880, opened 2025-01-10
Mr.Mulles German Translation

SnailPerson:
Don i have no idea if this is correct but this is how you told me to do it on discord


SnailPerson:
cancelled

PR #1885, opened 2025-01-24
Added more performant addLine for api purpose

KalypsoExists:
Please, I don't want to make a point3D when I directly have 6 floats!!!


DonBruce64:
Works for me!

PR #1897, opened 2025-02-22
Makes hasSkidSteer and MTA VMable

boot2big:
maxTiltAngle, hasSkidSteer and hasPermanentSkidSteer are now modifiable

If both HSS and HPSS are true, the latter will take precedence as the former will be disabled

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
@@ -73,6 +73,9 @@ abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding {
     public final ComputedVariable brakingFactorVar;
     public final ComputedVariable overSteerVar;
     public final ComputedVariable underSteerVar;
+    public final ComputedVariable maxTiltAngleVar;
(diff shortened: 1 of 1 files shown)


mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
```diff
@@ -522,14 +531,14 @@ private double getTurningForce() {
         double steeringAngle = getSteeringAngle() * 45;
 
         //Check for permanent skid-steer if we have it.
-        if (definition.motorized.hasPermanentSkidSteer) {
+        if (hasPermanentSkidSteerVar.isActive) {
```
DonBruce64, line 534: So I buy everything, except this.  I would think that being able to VM if skid steer is permanent essentially removes the "permanece" of the steer?  Like, you would set it permanent, but then a VM could yeet that to 0, and thus it wouldn't be permanent.

Am I making sense?  I get the feeling I only somewhat am.

boot2big: Sorry, hadn't been getting emails on Git updates, I really gotta fix that...

It is admittedly somewhat silly to have taken the _permanent_ version of a parameter and allow it to be made non-permanent, so I'll get that changed up quick-like.

PR #1903, opened 2025-04-18
New OCP Banner and Brand section (don't merge yet)

SnailPerson:
Experimental new approach to linking packs. One image speaks a thousand words, as they say...

This is reserved for the 1.20 launch, since i depicted vehicles that are not yet implemented into the CF version.

### DIFF
--- a/CurseSite.html
+++ b/CurseSite.html
@@ -9,9 +9,9 @@
 <h2><strong>Packs:</strong></h2>
 <p><strong>All vehicle packs go in your "mods" folder just like any other mod!</strong></p>
 <br>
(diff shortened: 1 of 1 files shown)


SnailPerson:
for context, this is how it looks like:

image

SnailPerson:
nope. OCP is not yet updated
ill let you know when its ready

DonBruce64:
@SnailPerson let's get this done in the next week or so.  I wanna clear out the PRs here.  Snail has had months for OCP, and then holds Don vigils when he's gone for months. ; )

PR #1904, opened 2025-04-19
Add "alwaysLit" instrument component definition

conman180:
I dont know why the other commits show up still. Also hello don hope all is well.


conman180:
Will add wiki entries once merged if approved

SnailPerson:
i second this

mccore/src/main/java/minecrafttransportsimulator/rendering/RenderInstrument.java
```diff
@@ -112,7 +112,7 @@ public static void drawInstrument(AEntityE_Interactable<?> entity, Transformatio
                         //Set points to the variables here and render them.
                         //If the shape is lit, disable lighting for blending.
                         renderable.setLightValue(entity.worldLightValue);
-                        if (component.lightUpTexture && lightsOn && ConfigSystem.client.renderingSettings.brightLights.value) {
+                        if ((component.lightUpTexture && lightsOn) || component.alwaysLit && ConfigSystem.client.renderingSettings.brightLights.value) {
```
DonBruce64, line 115: Can you add an extra () to put the new stuff into a group too?  I got confused reading this since I know that && and || have an order-of-operations and I thought they would behave differently.
((component.lightUpTexture && lightsOn) || component.alwaysLit)

DonBruce64 (changes requested):

conman180:
dammit it pushed multiple commits i changed but one PA approved one from before you left

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
```diff
@@ -455,7 +455,7 @@ protected void getForcesAndMotions() {
             }
 
             //Finally, get gravity.  Blimps sink when dead.
-            gravitationalForce = !ballastVolumeVar.isActive || outOfHealth ? currentMass * (9.8 / 400) : 0;
+            gravitationalForce = !ballastVolumeVar.isActive || outOfHealth ? currentMass * 0.0245 : 0;
```
DonBruce64, line 458: I get all the changes, but what is this guy about?  Not that it changes anything functionally, just more curious than anything.

SnailPerson: @DonBruce64 
image

DonBruce64: I get the fact it does that, but I was wondering why change it if it's working?  The compiler will see that and auto-complete it normally.  The reason I had it that way was for clarity as 9.8 is a gravitational constant, so it's easier to see what's going on there.  Again, I don't care one way or the other which it is in this case, twas for my own curiousness.  Will wait for Conman to get back from work.

conman180: its more my ocd and didnt realize that compiling simplifies it down in the final product. i figured this was just a wasteful operation taking an extra step where unneeded..

DonBruce64: That's a fair point.  I didn't realize it myself until I had to do some recovery work when I accidentally deleted some files rather than committing them.  When I got in there I realized that the compiler replaces any constants variable with hard-coded values, and any static math functions with their results.  It's why MC's source code looks a little funky with all sorts of factors with no names.

PR #1920, opened 2025-06-05
Electrical system changes & miscellaneous fixes

boot2big:
`electric_usage` is now VMable, functioning the same as it would if other sources were drawing from (or charging) the vehicle's battery.

Vehicle maximum battery voltage can now be redefined (and VMed) to allow for higher voltage electrical systems. 

When paired together, pack authors can balance these changes by multiplying `electric_usage` to make engine starters draw more (or less) power from higher (or lower) voltage systems.

Also fixed downForce not being LC'ed (should fix an odd pack here or there,) and Y2K button not shutting off lights along with the engine, and activating the parking brake.


boot2big:
Main inspiration for this PR was having additional power drains for vehicles, such as camper beds that have lights, radios/computers in emergency vehicles, etc.
Consequently, this allows for vehicles to charge themselves up as well.

This pull request will have a minimal effect on all vehicles from every pack, as it does technically increase the maximum voltage from 13V to 14V. Additionally all devmode exports will now feature a variable for battery power by default.

mccore/src/main/java/minecrafttransportsimulator/packloading/LegacyCompatSystem.java
```diff
@@ -50,7 +50,14 @@
 import minecrafttransportsimulator.jsondefs.JSONMuzzleGroup;
 import minecrafttransportsimulator.jsondefs.JSONPart;
 import minecrafttransportsimulator.jsondefs.JSONPart.CrafterComponentType;
-import minecrafttransportsimulator.jsondefs.JSONPart.EffectorComponentType;
+import minecrafttransportsimulator.jsondefs.JSONPart.EffectorCompon        //Set default health and battery capacity.
```
DonBruce64, line 53: I think this is some copy-pasta error?  Shouldn't be here.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleE_Powered.java
```diff
@@ -95,8 +97,17 @@ public AEntityVehicleE_Powered(AWrapperWorld world, IWrapperPlayer placingPlayer
...
     	addVariable(this.reverseThrustVar = new ComputedVariable(this, "reverser", data));
+    	addVariable(this.electricUsageVar = new ComputedVariable(this, "electric_usage", data));
+    	addVariable(this.batteryCapacityVar = new ComputedVariable(this, "batteryCapacity", data));
     }
-
+	
+    @Override
+    public void setVariableDefaults() {
+        super.setVariableDefaults();
+        electricUsageVar.setTo(electricUsage, false);
```
DonBruce64, line 107: I think you don't need the "electricUsage" variable anymore?  Things that reference this should be set to electricUsageVar.currentValue instead otherwise they won't see the changed value via VMs.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleE_Powered.java
```diff
@@ -79,7 +81,7 @@ public AEntityVehicleE_Powered(AWrapperWorld world, IWrapperPlayer placingPlayer
             this.selectedBeacon = NavBeacon.getByNameFromWorld(world, selectedBeaconName);
             this.fuelTank = new EntityFluidTank(world, data.getData("fuelTank"), definition.motorized.fuelCapacity);
         } else {
-            this.electricPower = 12;
+            this.electricPower = (definition.motorized.batteryCapacity * 0.857142857);
```
DonBruce64, line 84: I might make the fraction/decimal a constant up with the other public static final stuff, and make it 12D/14D so it's clear where this comes from vs just a decimal value.  Then if we ever decide to change it, it'll reflect in the code.

boot2big: As in, statically add 2 to `batteryCapacity`'s defined value? Just clarifying to be sure, but it can be changed to reflect future plans rather than having to change it later.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleE_Powered.java
boot2big: I can nix it, or maybe make it work-ish. Will try latter then former.

mccore/src/main/java/minecrafttransportsimulator/packloading/LegacyCompatSystem.java
boot2big: Oh crap, oughta fix that mebbe. Will do.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleE_Powered.java
DonBruce64: I mean making the fractional decimal a constant variable in the file like MAX_THROTTLE and such vs putting it in the code directly.  Then referencing that variable wherever we use the fraction.  That way if the fraction changes in the future or we want to do something with it, we only need to change it in one place.

mccore/src/main/java/minecrafttransportsimulator/packloading/LegacyCompatSystem.java
boot2big: Fixed, sorry it took so long... >.<

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleE_Powered.java
boot2big: I'm.. not sure how to implement this. I can't just nix it, and in testing I haven't been sure how it's been desynced by unseen changes? Sorry..

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleE_Powered.java
```diff
@@ -61,6 +63,7 @@ public abstract class AEntityVehicleE_Powered extends AEntityVehicleD_Moving {
     public String selectedBeaconName;
     public NavBeacon selectedBeacon;
     public final EntityFluidTank fuelTank;
+    public static final double batteryDefaultCharge = 0.85715D;
```
DonBruce64, line 66: Only final thing, constants are usually declared LIKE_THIS with all caps and underscores.  See MAX_THROTTLE up thar.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleE_Powered.java
LauraDarkez: @DonBruce64 can you clarify this? its been 2 weeks

mccore/src/main/java/minecrafttransportsimulator/items/instances/ItemItem.java
```diff
@@ -367,7 +367,7 @@ public CallbackType doEntityInteraction(AEntityE_Interactable<?> entity, Boundin
                 if (rightClick) {
                     if (entity instanceof EntityVehicleF_Physics) {
                         //Use jumper on vehicle.
-                        ((EntityVehicleF_Physics) entity).electricPower = 12;
+                        ((EntityVehicleF_Physics) entity).electricPower = (((EntityVehicleF_Physics) entity).batteryCapacityVar.currentValue * 0.857142857);
```
DonBruce64, line 370: I spy a missing default reference here....

DonBruce64 (changes requested):
Sorry, I missed one thing!

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleE_Powered.java
DonBruce64: By this I mean remove electricUsage as a variable.  Anywhere it's referenced, put electricUsageVar.currentValue, anywhere you set the value, do electricUsageVar.setTo or something similar.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartEngine.java
```diff
@@ -445,7 +445,7 @@ public void update() {
                 case NORMAL: {
                     if (running) {
                         //Provide electric power to the vehicle we're in.
-                        vehicleOn.electricUsage -= 0.05 * rpm / maxRPMVar.currentValue;
+                        vehicleOn.electricUsageVar.adjustBy(-0.05 * rpm / maxRPMVar.currentValue, true);
```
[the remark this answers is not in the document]
boot2big: I could've sworn I'd set this to true due to some issue I ran across, but it's possible that might've been something else... will compare soon to check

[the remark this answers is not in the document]
boot2big: Well sheee-oooot! I still don't remember what caused the desync issue before I smoothed that out, but it sure wasn't this. Thank you for pointing this out, Don! ^-^

DonBruce64:
Looks all good to me, I must be slipping or have looked at the old stuff...

boot2big:
Eehh!? Damnit GitHub... sure, can try to.

PR #1921, opened 2025-06-06
Add autopilot functionality

ktokto313:


mccore/src/main/java/minecrafttransportsimulator/guis/instances/GUIPanel.java
```diff
@@ -401,6 +401,44 @@ public void handleKeyTyped(char typedChar, int typedCode, TextBoxControlKey cont
...
+                                    //Update the vehicle beacon state.
+                                    try {
+                                        int heading = Integer.parseInt(getText());
+                                        InterfaceManager.packetInterface.sendToServer(new PacketEntityVariableSet(vehicle.autopilotHeading, heading));
+                                    } catch (Exception e) {}
+                                }
+                            };
+                            addComponent(beaconBox);
+                            newComponent = beaconBox;
+                            text = "Hdg";
```
DonBruce64, line 419: For all panel items, default text is in all CAPS.  Please change to match.

mccore/src/main/java/minecrafttransportsimulator/guis/instances/GUIPanel.java
```diff
@@ -401,6 +401,44 @@ public void handleKeyTyped(char typedChar, int typedCode, TextBoxControlKey cont
...
+                                    InterfaceManager.packetInterface.sendToServer(new PacketEntityVariableSet(vehicle.autopilotHeadingEnabledVar, 1));
+                                }
+                            }
+
+                            @Override
+                            public int getState() {
+                                return vehicle.autopilotValueVar.isActive ? 1 : 0;
+                            }
+                        };
+                        text = vehicle.definition.motorized.isAircraft ? "AUTO" : "CRUISE";
```
DonBruce64, line 439: I don't think this is right?  This is the stuff for the autopilot button, this wouldn't be the text for the heading box.

mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONPanel.java
```diff
@@ -121,6 +121,10 @@ public static enum SpecialComponent {
         CUSTOM_VARIABLE,
         @JSONDescription("Makes this a beacon box, which allows changing the selected beacon.  Will not be visible if hasRadioNav is false for the vehicle and allPlanesWithNav is false in the config.")
         BEACON_BOX,
+        @JSONDescription("Make this a heading input box, which allows changing the autopilot heading.")
+        HEADING,
+        @JSONDescription("Heading autopilot on or st")
```
DonBruce64, line 126: Change to clarify what "st" means.

mccore/src/main/java/minecrafttransportsimulator/packets/instances/PacketVehicleWaypointSelectRequest.java
```diff
@@ -0,0 +1,37 @@
+package minecrafttransportsimulator.packets.instances;
+
+import io.netty.buffer.ByteBuf;
+import minecrafttransportsimulator.entities.components.AEntityA_Base;
+import minecrafttransportsimulator.entities.instances.EntityVehicleF_Physics;
+import minecrafttransportsimulator.mcinterface.AWrapperWorld;
+import minecrafttransportsimulator.mcinterface.InterfaceManager;
+import minecrafttransportsimulator.packets.components.APacketEntity;
+
+public class PacketVehicleWaypointSelectRequest extends APacketEntity<EntityVehicleF_Physics> {
```
DonBruce64, line 10: This packet and the WaypointSelect packet can be one packet as they contain the same data, and just do different things based on if they are on the client or server.  Also, you're double-sending packets back.  Returning true on a packet sent to a server on the handle method will send that packet to all clients.  So you're sending the request packet back to do nothing with it, then sending the new select packet.  Just lump both together and return true and check for server/client for processing.

mccore/src/main/java/minecrafttransportsimulator/packets/instances/PacketVehicleWaypointUpdateRequest.java
```diff
@@ -0,0 +1,76 @@
...
+import io.netty.buffer.ByteBuf;
+import minecrafttransportsimulator.entities.instances.EntityVehicleF_Physics;
+import minecrafttransportsimulator.mcinterface.AWrapperWorld;
+import minecrafttransportsimulator.mcinterface.InterfaceManager;
+import minecrafttransportsimulator.packets.components.APacketEntity;
+
+/**
+ * Request waypoint selection update from client to server for vehicle
+ */
+public class PacketVehicleWaypointUpdateRequest extends APacketEntity<EntityVehicleF_Physics> {
```
DonBruce64, line 12: Same comment as prior waypoint packet set.

mccore/src/main/java/minecrafttransportsimulator/packets/instances/PacketWaypointUpdate.java
```diff
@@ -0,0 +1,65 @@
...
+
+import io.netty.buffer.ByteBuf;
+import minecrafttransportsimulator.baseclasses.NavWaypoint;
+import minecrafttransportsimulator.baseclasses.NavWaypointUpdater;
+import minecrafttransportsimulator.mcinterface.AWrapperWorld;
+import minecrafttransportsimulator.packets.components.APacketBase;
+/**
+ * Update waypoint from client to server
+ */
+public class PacketWaypointUpdate extends APacketBase {
```
DonBruce64, line 11: Same comment as prior waypoint packet set.

mccore/src/main/java/minecrafttransportsimulator/systems/ControlSystem.java
```diff
@@ -688,6 +706,7 @@ public enum ControlsKeyboard {
         GENERAL_CUSTOM3(ControlsJoystick.GENERAL_CUSTOM3, true, "NUMPAD2", LanguageSystem.INPUT_CUSTOM3),
         GENERAL_CUSTOM4(ControlsJoystick.GENERAL_CUSTOM4, true, "NUMPAD3", LanguageSystem.INPUT_CUSTOM4),
         GENERAL_RELOAD(ControlsJoystick.GENERAL_RELOAD, true, "R", LanguageSystem.INPUT_GUN_RELOAD),
+        GENERAL_WAYPOINT_MANAGER(ControlsJoystick.GENERAL_WAYPOINT_MANAGER, true, "5", LanguageSystem.INPUT_WAYPOINT_MANAGER),
```
DonBruce64, line 709: Don't use hotbar keys for defaults. They'll conflict with MC.

mccore/src/main/java/minecrafttransportsimulator/baseclasses/NavWaypoint.java
DonBruce64, line 1: So, rather than make a new thing in the way of waypoints, and a new GUI, is there any reason you can't just add these components to the existing beacons and have them work as waypoints?  They're already defined in the world and are designed for such things such as wayfinding.  I feel it's a more in-game way to handle things vs just menus and such.  Is this due to just not wanting to pack things into the GUI, or rather some whollistic approach?  I know in aviation there are difference between beacons and such, but for most players it's not a distinction they'll make.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleE_Powered.java
```diff
@@ -61,7 +59,14 @@ public abstract class AEntityVehicleE_Powered extends AEntityVehicleD_Moving {
     public double electricPower;
     public double electricFlow;
     public String selectedBeaconName;
+    public String selectedWaypointIndex;
     public NavBeacon selectedBeacon;
+    public final String SELECTED_WAYPOINT_LIST_KEY = "selectedWaypoints";
+    public final String WAYPOINT_ORDER_KEY = "WaypointOrder";
```
DonBruce64, line 65: Should be camelCase to match other variables.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleE_Powered.java
```diff
@@ -81,6 +86,28 @@ public AEntityVehicleE_Powered(AWrapperWorld world, IWrapperPlayer placingPlayer
...
             this.selectedBeacon = NavBeacon.getByNameFromWorld(world, selectedBeaconName);
             this.fuelTank = new EntityFluidTank(world, data.getData("fuelTank"), definition.motorized.fuelCapacity);
+
+            //init selectedWaypoints
+            try{
+                this.waypointData = data.getData(SELECTED_WAYPOINT_LIST_KEY);
+                this.selectedWaypointIndex = "-1";
+                if(this.waypointData != null){
+                    IWrapperNBT waypointOrderData = data.getData(WAYPOINT_ORDER_KEY);
+                    int ListIndex = 0;
```
DonBruce64, line 96: You don't use ListIndex in this loop?  I see what you're going for though, one should use the rolling index vs a getter, but is there any reason you need the index?  Just grab the keys via the "line" variable and get those string data.  The strings in waypointOrderData should be the keys of those waypoint data.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleE_Powered.java
```diff
@@ -301,6 +328,81 @@ public void updatePartList() {
...
+    //For Flight plan, selectedWaypoint should always be the last one of selectedWaypointList.
+    public List<NavWaypoint> selectedWaypointList = new ArrayList<>();
+    IWrapperNBT waypointData = InterfaceManager.coreInterface.getNewNBTWrapper();
+
+    //Update selected waypoint state
+    public void UpdateWaypointList(String operation,String opIndex,String index,String name,String targetSpeed,String bearing,String StrX,String StrY,String StrZ) {
+        int Index = -1;
+        int OpIndex = -1;
+        try{
+            Index = Integer.parseInt(index);
```
DonBruce64, line 342: No Capitalized variables in Java please.  Those are for classes.

mccore/src/main/java/minecrafttransportsimulator/baseclasses/NavWaypointUpdater.java
```diff
@@ -0,0 +1,102 @@
...
+                double z = Double.parseDouble(zStr);
+
+                if (currentWaypoint != null) {
+                    NavWaypoint.removeFromWorld(world, currentWaypoint.index);
+                    currentWaypoint = null;
+                }
+                currentWaypoint = new NavWaypoint(world,waypointIndex,waypointName,targetSpeed,bearing,new Point3D(x, y, z));
+
+            } catch (NumberFormatException e) {
+//            System.err.println("Invalid input: Expected numerical values for targetSpeed, bearing, and position.");
```
DonBruce64, line 50: No debug stuff.  Just leave the catch exception e and call it good.  Also, see prior comment about integration with beacons.

mccore/src/main/java/minecrafttransportsimulator/guis/instances/GUIWaypointManager.java
```diff
@@ -0,0 +1,689 @@
...
+        });
+    }
+
+    @Override
+    public void setupComponents() {
+        super.setupComponents();
+        //setupComponents will be executed every time the window is resized!
+
+        GUIComponentLabel GlobalWaypointLabel = new GUIComponentLabel(guiLeft + 20, guiTop + globalOffset, ColorRGB.WHITE, "Global Waypoints");
+        GUIComponentLabel V_GlobalWaypointLabel = new GUIComponentLabel(guiLeft + 20, guiTop + vehicleOffet, ColorRGB.WHITE, "F-Plan Waypoints");
```
DonBruce64, line 458: Variables shouldn't start with V_ in Java.  Also, camelCase for them please.  Normally IDEs like Eclipse and such yell at one about this.  Surprised yours didn't.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
```diff
@@ -66,6 +61,23 @@ public class EntityVehicleF_Physics extends AEntityVehicleE_Powered {
...
+    public final ComputedVariable autopilotSpeedEnabledVar;
+    public final ComputedVariable autopilotVerticalSpeedEnabledVar;
+    public final ComputedVariable autopilotPositionX;
+    public final ComputedVariable autopilotPositionY;
+    public final ComputedVariable autopilotPositionZ;
+    public final ComputedVariable autopilotHeading;
+    public final ComputedVariable autopilotAltitude;
+    public final ComputedVariable autopilotSpeed;
+    public final ComputedVariable autopilotVerticalSpeed;
+//    private PIDController verticalSpeedController = new PIDController(0.001, 0.000005, 0.01);
```
DonBruce64, line 76: No debug leftovers please.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
```diff
@@ -611,19 +635,65 @@ protected void adjustControlSurfaces() {
...
-            	elevatorTrimVar.adjustBy(0.1, true);
-            } else if (-motion.y * 10 < elevatorTrimVar.currentValue - 1 && elevatorTrimVar.currentValue > -MAX_ELEVATOR_TRIM) {
-            	elevatorTrimVar.adjustBy(-0.1, true);
-            }
-            //Keep the roll angle at 0.
-            if (-orientation.angles.z > aileronTrimVar.currentValue + 0.1 && aileronTrimVar.currentValue < MAX_AILERON_TRIM) {
-            	aileronTrimVar.adjustBy(0.1, true);
-            } else if (-orientation.angles.z < aileronTrimVar.currentValue - 0.1 && aileronTrimVar.currentValue > -MAX_AILERON_TRIM) {
-            	aileronTrimVar.adjustBy(-0.1, true);
+//            if (autopilotNavEnabledVar.isActive) {
```
DonBruce64, line 638: Debug leftover.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
```diff
@@ -611,19 +635,65 @@ protected void adjustControlSurfaces() {
...
+                setSpeed();
+            } else {
+                speedController.clear();
             }
+            if (autopilotVerticalSpeedEnabledVar.isActive) {
+                setVerticalSpeed();
+            } else {
+                verticalSpeedController.clear();
+            }
+//            if (selectedBeacon == null) {
```
DonBruce64, line 686: More leftovers.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
```diff
@@ -656,6 +726,159 @@ protected void adjustControlSurfaces() {
...
+
+        if (-orientation.angles.z > aileronTrimVar.currentValue + 0.1 && aileronTrimVar.currentValue < MAX_AILERON_TRIM) {
+            aileronTrimVar.adjustBy(0.1, true);
+        } else if (-orientation.angles.z < aileronTrimVar.currentValue - 0.1 && aileronTrimVar.currentValue > -MAX_AILERON_TRIM) {
+            aileronTrimVar.adjustBy(-0.1, true);
+        }
+    }
+
+    public void navGPS() {
+//        double heading = Math.toDegrees(Math.atan2(autopilotPositionZ.currentValue - position.z , autopilotPositionX.currentValue - position.x));
```
DonBruce64, line 745: Leftovers.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
```diff
@@ -656,6 +726,159 @@ protected void adjustControlSurfaces() {
...
+
+        // Vertical navigation
+        delta = selectedBeacon.getGlideSlopeDelta(this);
+        output = delta * 45;
+        if (output > 20) {
+            output = 20;
+        } else if (output < -45) {
+            output = -45;
+        }
+        // Output = Math.toDegrees(Math.asin(motion.y / velocity))
```
DonBruce64, line 777: Leftovers.

DonBruce64 (changes requested):

DonBruce64:
@ktokto313 In general, I agree with the changes here.  The allowing aircraft to follow waypoints.  The flight plan stuff, etc.  What gets me is that we're adding virtual waypoints to world data when we have a physical beacon thing that's already designed for such waypoints.  I feel that should be upgraded to have the same variables as the existing waypoints.  I do get that altitude and such isn't something that is normally defined by beacons, but I feel one could do that here to simply have people plan out routes for general flight this way.

Heading also isn't usually defined by single points, but getting it to follow a track wouldn't be hard.  Plus then one could have the plane switch to the next waypoint when it gets within 100 blocks or so of the active one.  Allows for automatic flying based on waypoint bits, and if we get glideslope to work could allow for automatic landing.

I do agree with the bones and maths and such, and I want to keep that.  Just want to be conscious of adding a bunch of different systems that interact differently when we already have something physical players can use and will probably understand better.  Sure it's not technically the same as a flight plan, but your average MC kiddie knows what height means and know how to click blocks and work with GUIs and such.

ktokto313:
The flight plan and virtual waypoint was @Deepseasaltyfish plan, I think he wanted to have a system where plane can follow waypoint without getting to the point manually, placing a beacon and setting it up. I haven't give it much thought as it's within the game design area itself. Personally, I would like to have to follow and change VOR beacon manually but it's only my two cent.

The glideslope already works but it's not really an autoland. Throttle autopilot doesn't work with throttle/hotas iirc. That's about the state of this PR right now.

Deepseasaltyfish:
after a few months i have some new ideas, for example we could have custom automatic controller algorithm instead of hard-coded PID controller.
also this PR has stopped for a long time and still was at a early state and still need much work, i think i will have more time the second half of the year

DonBruce64:
Marking this PR as draft since it's not ready for merge.

PR #1924, opened 2025-06-09
Mr.Mulles new Vehicle Charger

SnailPerson:
Old one looked like a toaster. Or a bomb. 
Not anymore, with Mr.Mulles™ new EV Charger™ PR™

Have a look!
image

Shes pretty, aint she!


boot2big:
Much better than my old charger model... to be fair I assumed it had to be one of those stupid machine-in-one-block models because it'd accept power from any side.
But frankly, I don't mind it being replaced one bit, especially with a top-quality Mulle Model:tm:! ..or Mulle-del?

PR #1928, opened 2025-06-13
EMERGENCY PR. fixes broken lang files.

conman180:
FIxes Snail's oversight on the new lang files. fixes syntax errors and duplicate entry keys that would crash game on startup.

 found issue while updating code to find game wouldnt load


SnailPerson:
did an oopsie, I missed some funny commas 

shame

PR #1932, opened 2025-07-04
autoReloads restored for handheld guns

boot2big:
If a gun is handheld and autoReloads, it won't wait for the player to press the trigger or press Reload before attempting to reload itself. Useful for guns that need a constant intake of ammo, should always be loaded, etc.

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
@@ -372,7 +372,7 @@ public class PartGun extends APart {
                 //Clipped guns we can't reload if we're already reloading.
                 if (!definition.gun.blockReloading && (!isReloading || (definition.gun.isClipless && reloadMainTimeRemaining != 0))) {
                     if (isHandHeld) {
-                        if ((loadedBulletCount == 0 && playerPressedTrigger) || isHandHeldGunReloadRequested) {
+                        if ((loadedBulletCount == 0 && playerPressedTrigger) || isHandHeldGunReloadRequested || definition.gun.autoReload) {
                             //Check the player's inventory for bullets.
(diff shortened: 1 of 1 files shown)


LauraDarkez:
This pr is very much appreciated, definitely hope it gets pulled

DonBruce64:
@LauraDarkez @boot2big please relay message to y'all and Snail that I'll get to this and be back in about a week.  Been.... crazy.... but I finally have light in my tunnel!

LauraDarkez:
no worries

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -860,7 +884,7 @@ private void handleControl(IWrapperEntity controller) {
             }
 
             //If we are holding the trigger, request to fire.
-            if (playerHoldingTrigger) {
```
DonBruce64, line 863: This is the only thing I'm worried about, this went from a variable to a CV, did you delete playerHoldingTrigger to make sure it wasn't used somewhere else?  I think this comes from packets.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
boot2big: I did not delete it from being a packet, but in my testing I haven't come across any issues when firing guns, be they vehicle-mounted or handguns.

Plus, with the way I've implemented this (packet controls the default CV state, which can be overridden), does it *need* to be removed? If so I can keep taking a crack at this, however it seems to work fine this way - do let me know if you're telling me this because it's inefficient, though.

DonBruce64:
@boot2big I think you might need to re-base this since it's "conflicting" due to the reloadDelay?

Issue #1933, opened 2025-07-09
Missile_Incoming crash


LauraDarkez:
[latest.log](https://github.com/user-attachments/files/21147525/latest.log)

DonBruce64:
Addressed in 22.18.0.

PR #1934, opened 2025-07-09
Fix missile_incoming and missle_#_func

boot2big:
Addresses #1933


LauraDarkez:
i hope this gets added soon. this has been a major bug for @Wartich, @Ilya-torshov and myself

PR #1938, opened 2025-07-22
Mr.Mulles Charger Icon Patch

SnailPerson:
Apparently the core item icon folder got nuked and I forgot to put the icons into each versions item texture repository. This patches that


DonBruce64:
Done.  Also @SnailPerson FYI I'm without any computers for the next week.  But I shall return.  And what stories I have!

SnailPerson:
@DonBruce64 
I went ahead and dropped the stable betas of 1.12 - 1.19 on CF as a bridge version until you get back to fix bugs for 22.19.
This is impotant to finally get a fix for all the bugs from 22.17 out so people stop reporting them and also for PAs to finally be able to update their packs that depend on features that are not part of 22.17 on CF.

Additionally, the PAs and I collectively decided to also upload 1.20 marked as a beta file. Its performance not even 24 hours after release is quite impressive, almost double as much s the popular 1.12 version. 
This is a great sign! See you next week, captain.

image

PR #1939, opened 2025-07-24
Add new GravityFactor definition

conman180:
Adds an opt in gravity modifier. Never shows up in a json unless intentionally put there. defaults to normal gravityFactor inside mtsconfig. I made this VMable to make more believable flight physics in my aircraft. Have had it tested by other PAs and have had positive responses to having it added.


mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
```diff
@@ -775,6 +776,8 @@ public ComputedVariable createComputedVariable(String variable, boolean createDe
                 return new ComputedVariable(this, variable, partialTicks -> beingFueled ? 1 : 0, false);
             case ("thrust"):
                 return new ComputedVariable(this, variable, partialTicks -> thrustForceValue, false);
+            case ("gravityFactor"):
```
DonBruce64, line 779: Not needed.  Adding it as an addVariable up top automatically puts it in the variable listing so you don't need to re-define it here.  It is, after all, a ComputedVariable, so you're creating a ComputedVariable for a variable.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
```diff
@@ -459,15 +462,13 @@ protected void getForcesAndMotions() {
...
             //Finally, get gravity.  Blimps sink when dead.
-            gravitationalForce = !ballastVolumeVar.isActive || outOfHealth ? currentMass * 0.0245 : 0;
+            gravitationalForce = !ballastVolumeVar.isActive || outOfHealth ? (currentMass * (0.0245D * (definition.motorized.gravityFactor != 0 ? gravityFactorVar.currentValue : ConfigSystem.settings.general.gravityFactor.value))) : 0;
+
             if (waterBallastFactorVar.isActive && world.isBlockLiquid(position)) {
                 gravitationalForce -= gravitationalForce * waterBallastFactorVar.currentValue;
                 elevatorTorque = -orientation.angles.x * 2;
                 aileronTorque = -orientation.angles.z * 2;
             }
-            if (!definition.motorized.isAircraft) {
```
DonBruce64, line 468: Removing this and moving logic up to the prior line makes the system apply gravity factor to everything.  That I get.  But what makes me question things is that now you're applying the gravity factor to planes since the default is the config system value.  Whereas before it wouldn't apply to them.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
```diff
@@ -459,15 +462,13 @@ protected void getForcesAndMotions() {
             }
 
             //Finally, get gravity.  Blimps sink when dead.
-            gravitationalForce = !ballastVolumeVar.isActive || outOfHealth ? currentMass * 0.0245 : 0;
+            gravitationalForce = !ballastVolumeVar.isActive || outOfHealth ? (currentMass * (0.0245D * (definition.motorized.gravityFactor != 0 ? gravityFactorVar.currentValue : ConfigSystem.settings.general.gravityFactor.value))) : 0;
```
DonBruce64, line 465: Looking at this, I get what you're trying to do, but it's a bit hard to read.  How about this:
When you set the default value up in the defaults,
check if the motorized gravity factor is set.  If so, use that.
If not set, check if not an aircraft.  If not an aircraft, use the config value.

This lets you use the motorized value for anything, keeps only cars by default using the config value, and still lets you change the value via VMs post-vehicle creation.

Thoughts?

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityVehicleF_Physics.java
conman180: pushed change requested here

PR #1951, opened 2025-08-15
Fixed the use of a ticket that put the player in a vehicle instead of mobs

dldev32:
I removed the condition that the seat is a controller from the code. This is needed for future AI features, especially AI driving and one-seated vehicles.


dldev32:
> Again though, how does this work if you want to transport a villager? He'll pick the first seat he sees, which will likely be the driver's seat.

What if the vehicle is single-seated? However, there are many packs that have a controller seat linked to the guns, and so the previous piece of code might not allow a mob to be placed there, instead the AI rider would occupy any other seat that isnt a controller.

DonBruce64:
Noted.  Final check: do mobs end up in the seats with tickets?  I notice you removed the IMob check when you replaced it with EntityPlayer.

DonBruce64:
We don't want hostile mobs in the seats.  They reasonably would never be able to be put into a seat.  Especially since you could shift-click to load up a bunch of villagers in a 15 block radius and accidentally load a zombie or creeper underground. Please fix.

PR #1960, opened 2025-08-21
Electrical changes redux

boot2big:
Feck you GH...

original:

`electric_usage` is now VMable, functioning the same as it would if other sources were drawing from (or charging) the vehicle's battery.

Vehicle maximum battery voltage can now be redefined (and VMed) to allow for higher voltage electrical systems. 

When paired together, pack authors can balance these changes by multiplying `electric_usage` to make engine starters draw more (or less) power from higher (or lower) voltage systems.

Also fixed downForce not being LC'ed (should fix an odd pack here or there,) and Y2K button not shutting off lights along with the engine, and activating the parking brake.


boot2big:
Re-requested pull sponsored by **RED SHADOW MAGGOTS**

mccore/src/main/java/minecrafttransportsimulator/packloading/LegacyCompatSystem.java
```diff
@@ -360,11 +361,14 @@ private static void performVehicleLegacyCompats(JSONVehicle definition) {
...
 
-        //Set default health.
+        //Set default health and battery capacity.
         if (definition.general.health == 0) {
             definition.general.health = 100;
         }
 
+        if (definition.motorized.batteryCapacity == 0) {
+            definition.motorized.batteryCapacity = 14;
+
```
DonBruce64, line 371: One minor thing, are we missing a } here?

DonBruce64 (changes requested):

DonBruce64 (approved):

PR #1963, opened 2025-08-25
Explosivefix

maxizxyz:
Fixed explosions (which now deal splash damage)


dldev32:
I have tested your code before, but it sometimes crashes the game. See details here: https://discord.com/channels/232316230852280320/1403763899119173653

maxizxyz:
> The question I have is why is this needed? Shouldn't the player be affected by explosions by default? Why are we manually hurting them?

By default, explosions do not cause splash damage to players, only to mobs

DonBruce64:
> > The question I have is why is this needed? Shouldn't the player be affected by explosions by default? Why are we manually hurting them?
> 
> By default, explosions do not cause splash damage to players, only to mobs

This doesn't make sense though, I know if I set off a TNT, the player will get hurt.  You say they don't cause damage to players, is this only explosions from MTS, or from Minecraft in general?  If it's only one of them, I'm concerned about this since it should behave the same no matter which explosion happens.

PR #1966, opened 2025-08-26
Mr.Mulles new Jerrycan

SnailPerson:
New Jerrycan. wohoo
Look how snazzy. Yes. Yay. Christmas!


SnailPerson:
yes but that would save you around lemme check... 150 bytes lmao
image

Trust me that's not worth redoing the model for UVing and reuploading the entire thing on git times 5.

Please merge.

PR #2022, opened 2026-01-31
Mr.Mulles snowfall_strength

SnailPerson:
Daddy don please verify these I hope MTS doestn die, It should work, it works on my end, perhaps compile a version and send it to me on discord so we can make sure I didn't write big poopoo here. this is my first code contribution to MTS so please be gentle


SnailPerson:
daddy @DonBruce64 please have a looky

SnailPerson:
piss piss made a better PR

PR #2026, opened 2026-02-08
Ext tank match fuel loader flowrate

Hardwarkid:
Discussion revolving around this PR is here:
https://discord.com/channels/232316230852280320/1466226017780764851

I have been testing for several days, at maxAmount = 100 and that works fine, so I assume this will as well.


SnailPerson:
@DonBruce64 
I had this conversation with Hardwarkid.

Initally, the problem stemmed form engines with high fuel consumption consuming faster than the external tank can resupply, leading to spammed engine stall and engine start sounds. A flow rate of 10 mB per tick (5 seconds per bucket instead of 50 seconds) seemed to be a good middle ground

PR #2031, opened 2026-02-18
Mr.Mulles TPP freecam config

SnailPerson:
@DonBruce64 Disconnects third person perspective camera from vehicle orientation, when Freecam_3P is true

https://github.com/user-attachments/assets/609311c7-fdd5-4480-be4e-81707dfc5b95


DonBruce64:
@SnailPerson you're good with this being a P key config vs a keybind, right?

SnailPerson:
there was a keybind? it never showed up in the normal menus though... which one was it and how could I "reenable" it?

PR #2037, opened 2026-02-23
Mr.Mulles Vehicle Voiding Fix

SnailPerson:
@DonBruce64 

I spent the better part of today tinkering with MTS. reading through .java files, asking AI for explanations (they're great teachers for all things code, I must admit) and engineered a solution for the voiding issue.

Instead of acting when the vehicle already is inside an unloaded chunk, we prevent it form even getting there.
There's a guard / checker always on the vector of a vehicle, dynamically checking the chunks up ahead and confirming wether they're loaded or not.

If not, the speed of the vehicle is immediately throttled down to prevent it from going into unloaded chunks at all.
If the chunk is loaded, the speed cap is removed and the vehicle is free to accelerate again.
Tested this in an amplified world with very little RAM and the results were very promising so far. 
Eventhough I had a cracked up jet with huge engine power, I was not able to go faster than the chunk loading speed.

Please have a look if there's anything I can optimise


DonBruce64:
Converting this to draft since you and Nax both have a PR out for this and I've sent comments on Nax's.  I don't want to close this since you might need this for code reference.

PR #2048, opened 2026-03-02
Fix Vehicle entering unloaded chunk to prevent game kicking player off vehicle on MC1165 - MC1201

Naxgeneral:
**About**
This is a fix that prevent vehicle from entering unloaded chunk. 

**How it work**
1. Get a block sample by using vehicle's point 3D X,Z position with fixed Y pos being 1. added by vehicle motionApplied * 42 to get position infront of vehicle based on motion. 

2. Then cache it in quarantine zone to check if it loaded or not by using minecraft's internal function method.
 
3. While this being checked it set vehicle's speed factor to 0.009 so vehicle is slow enough to wait for chunk to load. while also set timer to check if player turn away from this unloaded chunk or not to prevent vehicle always travel at 0.009 SF while turn away from unloaded area.

4. If the chunk in the quarantine zone is loaded vehicle will be let go and using its own respective speed factor and get new block sample.

**Advantage :**
This method will not force limit vehicle actual movement speed which will not cause some plane that require high speed to stall.
**Disadvantage :**
~~This probably caused by minecraft basic function stop working when entity left build height cause it seem like the system just not work at all beyond build height~~  This should be fixed in latest commit

**My own Test result :**
**Client :** 
I tested this my self on my mod pack setup that somehow make world generate 1 chunk per 2 business day and it work as intended
**Server :**
as well as in server environment  i usually have bad connection between a server (800ms ping) which usually make server not realize my actual position and not load chunk in the path and make game kick me out from plane but with this fix i able to fly even with bad internet connection or  during high server load.

foot note : 
I have to test this all my self as i posted beta and 9999 people who crying over this issue just vanished like they vaporized and not even come to test the fix.

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
@@ -44,11 +44,11 @@ import minecrafttransportsimulator.systems.LanguageSystem;
  * @author don_bruce
  */
 abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding {
-	//Variables
-	public final ComputedVariable leftTurnLightVar;
-	public final ComputedVariable rightTurnLightVar;
-	public final ComputedVariable brakeVar;
-	public final ComputedVariable parkingBrakeVar;
+    //Variables
+    public final ComputedVariable leftTurnLightVar;
+    public final ComputedVariable rightTurnLightVar;
+    public final ComputedVariable brakeVar;
+    public final ComputedVariable parkingBrakeVar;
     public final ComputedVariable lockedVar;
     public static final double MAX_BRAKE = 1D;
     public UUID keyUUID;
@@ -58,6 +58,9 @@ abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding {
     public boolean slipping;
     public boolean skidSteerActive;
     public boolean lockedOnRoad;
+    public double modifiedSpeedFactor = speedFactor;
+    public boolean chunkCacheStat;
+    public int cachedChunkTimeout = 0;
     private boolean updateGroundDevicesRequest;
     private int lastBlockCollisionBoxesCount;
     private int crashDebounce;
@@ -113,6 +116,8 @@ abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding {
     private final Point3D tempBoxPosition = new Point3D();
     private final Point3D normalizedGroundVelocityVector = new Point3D();
     private final Point3D normalizedGroundHeadingVector = new Point3D();
+    private final Point3D chunkGuardDirection = new Point3D();
+    private final Point3D cachedChunkGuardDirection = new Point3D();
     public final List<BoundingBox> allBlockCollisionBoxes = new ArrayList<>(); //Public so we can add ground device boxes to this set.
     private AEntityE_Interactable<?> lastCollidedEntity;
     public VehicleGroundDeviceCollection groundDeviceCollective;
@@ -137,7 +142,7 @@ abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding {
         this.groundDeviceCollective = new VehicleGroundDeviceCollection((EntityVehicleF_Physics) this);
         this.placingPlayer = placingPlayer;
         this.blockBreakDelay = 500;
-        
+
         addVariable(this.leftTurnLightVar = new ComputedVariable(this, "left_turn_signal", data));
         addVariable(this.rightTurnLightVar = new ComputedVariable(this, "right_turn_signal", data));
         addVariable(this.brakeVar = new ComputedVariable(this, "brake", data));
@@ -223,6 +228,7 @@ abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding {
             } else {
                 slipping = false;
             }
+            modifySpeedFactor();
             world.beginProfiling("TotalMovement", false);
             moveVehicle();
             if (!world.isClient()) {
@@ -236,7 +242,6 @@ abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding {
     @Override
     public void doPostUpdateLogic() {
         super.doPostUpdateLogic();
-
         //Move all entities that are touching this entity.
         if (velocity != 0) {
             world.beginProfiling("MoveAlongEntities", true);
@@ -598,7 +603,7 @@ abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding {
                 //This is because the faster we go the quicker we need to turn to keep pace with the vehicle's movement.
                 //We need to take speed-factor into account here, as that will make us move different lengths per tick.
                 //Finally, we need to reduce this by a constant to get "proper" force..
-                return turningForce * groundVelocity * (speedFactor / 0.35D) / 2D;
+                return turningForce * groundVelocity * (modifiedSpeedFactor / 0.35D) / 2D;
             }
         }
         return 0;
@@ -787,11 +792,11 @@ abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding {
                 //whichever is the lower of the two.  If we apply boost, update our collision boxes before the next step.
                 //Note that this logic is not applied on trailers, as they use special checks with only rotations for movement.
                 world.beginProfiling("GroundBoostCheck", false);
-                groundMotion.y = groundDeviceCollective.getMaxCollisionDepth() / speedFactor;
+                groundMotion.y = groundDeviceCollective.getMaxCollisionDepth() / modifiedSpeedFactor;
                 if (groundMotion.y > 0) {
                     world.beginProfiling("GroundBoostApply", false);
                     //Make sure boost doesn't exceed the config value.
-                    groundMotion.y = Math.min(groundMotion.y, climbSpeedVar.currentValue / speedFactor);
+                    groundMotion.y = Math.min(groundMotion.y, climbSpeedVar.currentValue / modifiedSpeedFactor);
 
                     //If adding our boost would make motion.y positive, set motion.y to zero and apply the remaining boost.
                     //This is done as it's clear motion.y is just moving the vehicle into the ground.
@@ -890,7 +895,7 @@ abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding {
 
             //Now that that the movement has been checked, move the vehicle.
             world.beginProfiling("ApplyMotions", false);
-            motionApplied.set(motion).scale(speedFactor).add(groundMotion);
+            motionApplied.set(motion).scale(modifiedSpeedFactor).add(groundMotion);
             rotationApplied.angles.set(rotation.angles);
 
             //Add road contributions.
@@ -903,7 +908,7 @@ abstract class AEntityVehicleD_Moving extends AEntityVehicleC_Colliding {
                         pathingApplied = -pathingApplied;
                     }
                 } else {
-                    pathingApplied = goingInReverse ? -velocity * speedFactor : velocity * speedFactor;
+                    pathingApplied = goingInReverse ? -velocity * modifiedSpeedFactor : velocity * modifiedSpeedFactor;
                 }
             } else {
                 pathingApplied = 0;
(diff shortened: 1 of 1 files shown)


Naxgeneral:
Additional comment :
I might also make it configurable through mtsconfig.json as in older version  such as 1.12.2 there no need to do this. as well as temporary speed factor it change to. 

This base code does fix the issue but the method that get block sample to throw into minecraft's chunk checking function just ditch it self beyond build height so if there suggestion or better method for this i also can try it .

-# This code was wrote all by my self with some research of basic Minecraft function through documentation that so hard to find so i know everything i wrote.

SnailPerson:
@DonBruce64 So we have an issue here, both me and nax have a running attempt at fixing this issue. 
My system uses a point 3d to determine whether a chunk upfront (vector of the vehicle) is loaded or not, which scales dynamically with speed. Should the point detect an unloaded chunk it will slow down the plane temporarily until the chunk up ahead is loaded. Critical difference is that it works above build height.

Max system works as described above just that the "dealbreaker" is the system no longer working above build height, which is unfortunately also the height most players will fly above of rather than below. Do you have any insight how we can get the best of both worlds?

DonBruce64:
@Naxgeneral @SnailPerson so looking at both your codes, I can see what the idea is.  Slow down the vehicle when it gets near unloaded chunks.  However, you're modifying the speed factor only in some parts of the code.  And also you're doing this modification on the client and server, which has the problem of causing de-syncs if the client takes longer to load a chunk than the server, and thus gets knecapped sooner.

What I would potentially try is the following:
Rather than slow down the vehicle, instead, have it not update.  So in EntityManager check if a entity will move into an unloaded chunk with its current speed, plus a small safety factor of perhaps 1.25.  If it will move into an unloaded chunk with this speed, skip updates.

With this, the vehicle should simply halt processing on the server once it gets to an unloaded chunk.  The client could run this code too, and since both don't adjust speed factor, the internal code will act as if it just didn't get a tick update and won't know anything is outside the ordinary.  This also ensures that syncing operations don't get mal-adjusted, as they are designed to dynamically rubber-band assuming the server and client have the same math applied.  And this is the key: if you change the speed factor on the server for 2 ticks, but the client for 8, then they are going to get out of sync due to them having different physics calculations.  But if you block the server for 2 ticks and the client for 8, that just means the client is "lagging" from the server with different calculations and will "catch-up".  In the flip side, it might get "pushed back" by the syncing code.  But the root thing here is that you don't dyanmically change the physics maths which are fairly important and need to run at a consistent settings in order to prevent issues.

To effect this, you'll probably want to leverage similar logic to the towing vehicles.  Having a bit that notes updates are skipped, as you'll need to not only block the vehicle's updates, but ensure that any parts that are on the vehicle also don't process updates.  These parts can just check if the vehicle update was blocked, via a function call.

So something like this in EntityManager
if entity not instanceof vehicle or entity.isnotgoingintounloadedchunk() then
update entity
entity.blockedupdate = false;
else
entity.blockedUpdate = true;
endif

Then in the vehicle
public boolean entityisnotgoingintounloadedchunk()
Logic to check this based on physics

Then in the part
Override
public boolean  entityisnotgoingintounloadedchunk()
vehicleon == null || vehicleOn !blockedUpdate

Wartich:
I've got a question about this, what if one goes a different way about it? If i remember it right, EObeta had the exact same issue in the past, how did you fix it without slowing the planes down?

DonBruce64:
EOBeta worked by making the planes not tied to any MC entity.  They existed in their own system.  That's easy enough to do.  What's hard and I never could get working is having the players properly ride the planes.  I could constantly warp the player to the plane and adjust their angles, but that caused issues with them jittering around on servers due to them not being in loaded chunks.  Not to mention a lot of the mods that didn't see the players as riding something and didn't adjust their behavior accordingly.  Finally, there were a lot of issues with trying to handle collision checks while out-of-bounds.  Planes falling through not-yet-loaded blocks was a constant issue.

It could maybe be done that way if one took a crack at it, but it would be a tough challenge.

DonBruce64:
Converting this PR to draft since there are systematic changes required and it's not ready for implementation yet.

Naxgeneral:
> > oh wait or do i have to do something like EntityVehicleF_Physic entity = new EntityVehicleF_Physic(args); and reference it off entity like that
> 
> No, using new makes a new vehicle. You don't want that. You need to reference the existing entity. Show code that's not working in picture and I can give you the proper formatting. Something like EntityVehicleF_Physics vehicle = (EntityVehicleF_Physics) entityObject;

@DonBruce64  
image
this is what i trying to do last time but not working as i realized that in this method it self the 'entity' is referred to AEntityA_Base which didnt inherited the Physic class so i cant referenced to the method i created in Entity Moving class so do you have idea what i can do here?

very late reply cause i was busy that time

DonBruce64:
@Naxgeneral Ahh, you want it slightly different.  You want to check if the entity is a definable entity, and make sure if it is, it won't tick in an unloaded chunk.  for that, you want this.
```
        if (entity instanceof AEntityD_Definable) {
            AEntityD_Definable<?> definable = (AEntityD_Definable<?>) entity;
            if(definable.notInUnloadedChunk()){
                entity.world.beginProfiling("MainUpdate", false);
                entity.update();
                entity.world.beginProfiling("PostUpdate", false);
                definable.doPostUpdateLogic();
                entity.world.endProfiling();
            }
        } else {
            entity.update();
        }
```

Until you tell Java that the entity is actually a definable, it won't know.  that's why the definable is used there down below.

Naxgeneral:
@DonBruce64  i tried your suggestion but it still can not reference to 
 the "notInUnloadedChunk" method i created inside EntityVehicleD_moving which is inherited to EntityF_Physic and the EntityD_Definable is clearly imported EntityF to use its properties 
 
 image

DonBruce64:
@Naxgeneral EntityVehicleD_moving is for vehicles.  You should put that method in EntityD_Definable, or better yet, EntityB_Existing.  Since it's a method that might get used for other entity types such as placed parts and such.  Then it'll work.

Naxgeneral:
@DonBruce64 I didnt know that i could just put my function in those classes, the issue is now fixed but one thing left was Part are kinda lagging behind when vehicle got stop ticking but from what i see i think it because of this also stop the part individually one last question was is there a way to detect if that definable is a part so i make it bypass this and obey vehicle tick?

https://github.com/user-attachments/assets/f7c0cf4c-378d-477f-ab0a-e94fb08f3146

DonBruce64:
@Naxgeneral  yup.  What you need to do is have the part `@Override` the function you defined.  And inside the function call entityOn.  That will make the part check if it's in an unloaded chunk by checking the parent vs itself.  So it'll all link to a vehicle.

You also might save the state of the function and only update the logic if the entity tick number changes.  Cause if you have a vehcile with 20 parts, they all will call the vehicle's loaded chunk function and check chunks.  That's a bit laggy.  You probably should save the value of that function in a boolean, and if you see the ticksExisted is different than the last time the function was called, then update the saved value.

```
@Override 
public boolean notInUnloadedChunk(){
    entityOn.notInUnloadedChunk();
}
```

Naxgeneral:
@DonBruce64 sorry for last false ping i forgot to add checking statment but i added the override method you suggested to the APart class and it stopped doing wobbling but the issue now only present with particle and entity that sitting on seat part instead 
image

image

Naxgeneral:
or maybe is this good enough for preventing vehicle from travelling into unloaded chunk
since othermod also didnt have very prestige method unless if they are VS or AeroNautic

DonBruce64:
Er, no.  I think you have that code confused.  Show the code you put in for the B_Existing class?

Naxgeneral:
@DonBruce64  here the comment is kinda outdated since it the same one that used for modifying speed factor



     public boolean isChunkLoaded(){

        //get a sample of block blocks in front of vehicle
        chunkGuardDirection.set(position.x + getMotionX(), 1D, position.z + getMotionZ());

        //check for stat 1 is cached unloaded chunk present , 0 is no cached chunks.
        if(!chunkCacheStat){
            //check the chunk sample is it not loaded if it is not loaded set vehicle speed factor to 0.008
            // if it is loaded make it follow its respective speed factor.
            if(!world.chunkLoaded(chunkGuardDirection)){
                chunkCacheStat = true;//tell them that there unloaded chunk cached so it will not do this again
                cachedChunkGuardDirection.set(chunkGuardDirection); //throw sample into quarantine zone to monitor it.
                cachedChunkTimeout = 0; //set timer in case player turn around.
                return false; // set speed factor
            }
            else {
                return true; // in case it stuck in this if.
            }
        }
        else{
            if(world.chunkLoaded(cachedChunkGuardDirection)){//check the quarantine zone does this chunk sample finally loaded
                chunkCacheStat = false;// tell them that there no more chunk in quarantine
                return true; // set speed factor back to default if it is loaded
            }
            else if(cachedChunkTimeout == 1000){//check if cycle already run for 1000 tick
                chunkCacheStat = true;
                cachedChunkTimeout = 0;// if yes tell chunk cache to stop and do cycle above again to get new sample which used incase player turn around to loaded chunks
            }
            else {
                cachedChunkTimeout++; // +1 to timer each cycle
                return false; // in case it fell off the chair.
            }
        }
        return true;
    }

    private double getMotionX(){
        if((motion.x * 32) > 64){
            return 64;
        }
        else if((motion.x * 32) < -64){
            return -64;
        }
        else {
            return motion.x * 32;
        }
    }

    private double getMotionZ(){
        if((motion.z * 32) > 64){
            return 64;
        }
        else if((motion.z * 32) < -64){
            return -64;
        }
        else {
            return motion.z * 32;
        }
    }`

SnailPerson:
@DonBruce64 just pinging you to bump, im not sure if you still get notifications from the gun GUI thread, but please have a look at that PR. Everything should be functional

DonBruce64:
@SnailPerson #2141 you mean?  I don't see any new comments there from anyone.

DonBruce64:
@Naxgeneral so for that function there, you should have it only run if the ticksExisted > lastTicksExisted.  If so, run the function and update the last value for the entity being in a chunk.  So basically, you're running the chunk-checking code only if the entity has ticked.  Then you're saving that value in a boolean and using that value if the ticks haven't changed.  That prevents the need to re-run the chunk checking code each tick, say  if each part also checks if the vehicle is in the chunk.  Er, does that make sense?  @SnailPerson  you might have better luck explaining in Discord perhaps, or perhaps boot to big might also understand what I'm saying.

Naxgeneral:
@DonBruce64  isnt making  whole function only probe and check chunk only when vehicle is ticking while saving last value that it being in unloaded chunk will make it stuck in softlock as it make vehicle no longer check if that last chunk it found is loaded or not. or i understand this wrong

maybe there something in vehicle that doesnt obey its own tick update
and it look like that there something that doesnt obey the doTick and still try pushing player into unloaded chunks anyways as the player keep getting pushing forwards aswell as particles but because of main body of vehicle got stopped behind so they got teleport backwards then continue to their last position

SnailPerson:
@DonBruce64 nay, I meant https://github.com/DonBruce64/MinecraftTransportSimulator/pull/2089

DonBruce64:
@Naxgeneral Ah, you are correct about that thought.  So, what you need is a boolean parameter on that funciton.  `useLastCheckValue`.  When you check in the EntityManger, you pass it false to force it to check and set the saved value.  If you're calling that function from parts, then it should be true so it uses the last checked state, if one is present.  That will force a check on the time the entity is ticked, but will ensure that last state is used for parts.

For particles, that's because the spawnParticles code is being updated on the rendering routine, but needs to not be updated if the thing is in unloaded chunks.  You should have the code that calls that function not do so if the entity that is spawning the particles isn't in a loaded chunk.

As to player, that I'm not sure about.  You say when the vehicle stops the player keeps moving forwards?  Or more they get pushed into the unloaded chunk while the vehicle stays behind?  Or do you mean they go into the unloaded chunk and then they get warped back into the loaded chunk?

Naxgeneral:
@DonBruce64 
Player and particle look like they still try to continue moving with last vehicle motion into unloaded chunk even whole vehicle got its tick stop so it look like player got wobble forwards and teleport back into seat but each time they shoot forward they go further and finally got banished by minecraft's very prestige system

But i think on part side it should be the same as checking if its parent still ticking? cause i make it return true when parent ticksexisted is more than its lastticksexisted

DonBruce64:
I think you're right, that will work for parts because you can check if their parent got an update.  Just re-name the variable to entityOnLastTicksExisted to avoid confusion as to the use-case.

As to the particle, you mean it spawns particles or they keep trying to move with the vehicle and the game gets confused as to their movement?

Player-wise that's confusing.  Is the player's seat entity moving?  It shouldn't be, it gets set to the entity position.  Can you cross-check that just in case?

https://github.com/DonBruce64/MinecraftTransportSimulator/blob/cd9cfb8fe74dbcc426eb830f14822adf7402261f/mcinterfaceforge1201/src/main/java/mcinterface1201/BuilderEntityLinkedSeat.java#L61

PR #2054, opened 2026-03-10
Mr.Mulles Tile Entity Fix

SnailPerson:
Fixes tile entities disappearing. Problem was   
this fucker.

image

apparently tile entities are failing at the exact same tick and then disappear.
On failure, MTS currently deletes the block (level.removeBlock(...)) in BuilderTileEntity.java

the thrown error is exactly from missing pack item (null item -> generateDefaultDefinition() throw) in
AEntityD_Definable.java and AEntityD_Definable.java.

affected Forge builders were not actually preserving lastLoadedNBT when saved before
first tick (tag = lastLoadedNBT in save path), causing NBT loss

Will also put into 1.21

### DIFF
--- a/mcinterfaceforge1201/src/main/java/mcinterface1201/BuilderTileEntity.java
+++ b/mcinterfaceforge1201/src/main/java/mcinterface1201/BuilderTileEntity.java
@@ -190,7 +190,7 @@ public class BuilderTileEntity extends BlockEntity {
             //Need to have this here as some mods will load us from NBT and then save us back
             //without ticking.  This causes data loss if we don't merge the last loaded NBT tag.
             //If we did tick, then the last loaded will be null and this doesn't apply.
-            tag = lastLoadedNBT;
+            tag.merge(lastLoadedNBT);
         }
     }
 }


DonBruce64:
I swear Mulle, it's shocking how much coding has changed since I started up this project, and how good AI has gotten for debuggering these things.  One of these days I need to grab a glass O something and just have a VC with you about stuff. on code and whatnot.  Feels like I kinda dropped the keys in your lap without much fanfare and yet you've fixed two of the most annoying bugs I never could figure out.

Is this what getting old feels like?  Lol, joking, but there's a wee bit O truth in there even with that.

SnailPerson:
in all honesty, im split on the use of AI. It is undoubtedly a revolutionary tool, especially in software development. It opens up the door for many artists or aspiring programmers to do things no-one imagined before, doing work in an afternoon a team woud've done in a week. 
A powerful tool, and a dangerous one too. It's the golden age of AI, and things seem nice for now but it's worrisome what it can be used for. Just like the discovery of nuclear energy brought us clean energy on one hand, and the bomb on the other.
Artificial intelligence is a great tool but an inexplicably great deception and manipulation system aswell, being able to conjure audio and video of things or people that never existed or make people look like they said things they never did.
And I cannot imagine what an AI means to a government which has postponed its morals. Not just because of the heap of personal information and data that can be gathered about each individual person, but also an open window in influencing the opinions of a wide spread of users by biasing outputs. It's scary, even for younger folks like me.
And on a more superficial level, im aware im able to do incredible things with MTS, fix bugs in record time and do things 5 times as quick. But im aware that this simplicity comes at the price of not really using your own brain that much. You command something to help you, more than you actually sit down and build these accomplishments from ground up. You dont learn by yourself. The deep knowledge you have is something I may be able to substitute by the use of AI, but never achieve with it. 
It's why im conflicted with using it at all. It helps me accomplish great things I wouldn't be able to do all by mylself, but the cost of running on mechanical legs is the atrophy of your own.

This is quite the novel ive written here haha... but yeah, id be up for a drink ;D

DonBruce64:
I think that mechanical legs are good in the right spots.  Take for instance modeling.  I made the MC-172 via raw programming.  It took me about a week to do.  Something that someone with good knowledge of Blender can do in a day.

And look at how folks created the first mario.  Using things like graph paper and such.  Nowdays you can just drag n drop.

I feel that each tool comes with pluses and minuses.  One can make bad models with Blender because it enables one to make models where they couldn't.  one can make bad mario maker stages when before they couldn't.  And one can now do so with code.  But I don't consider that to be bad, just an evolution of tools.

True, there offers avenues for manipulation, but that I feel would have happened with or without AI.  Just different things.  But it does indeed change how the world works and sees things.  With Plankton singing Country Roads comes Deepfakes.

I think that the world has indeed embraced technology without fully understanding the implacations.  But I suppose that happens with all things science.  Just look at Asbestos, or CFCs.  Things we thought were good and safe but had a dark side.

Do I think it's all sunshine?  No.  But I belive that worrying today empties tomorrow of it's joy.  And why make oneself sad about something one is controlling the best they can via voting and personal choices?

Ah, but looking me getting all whimsical in a PR.  Tis a thing for a Discord chat... I really should pop on there once in a while...

PR #2055, opened 2026-03-10
Default lockon type improvement

Wartich:
Makes so that default lockon can use distance and cone angle, just like boresight

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
@@ -150,8 +150,6 @@ public class PartGun extends APart {
     private final Point3D normalizedEntityVector = new Point3D();
 
     //Global data.
-    private static final int RAYTRACE_DISTANCE = 750;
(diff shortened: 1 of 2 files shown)


DonBruce64 (changes requested):

DonBruce64:
@Wartich 
`minecrafttransportsimulator.entities.instances.EntityVehicleF_Physics.createComputedVariable(EntityVehicleF_Physics.java:845)`

Check that line in that class.  It should tell you what computed variable it doesn't like that it's trying to make.  Probably something to do with radar or something.

Wartich:
Alright crash report was my stupidity. I tested in my "master" branch with other stuff piled in, and vehicle had missile_lockedonto variable on it, so when i spawned that vehicle in this PR branch it crashed since there's no such thing. Removed the unused stuff in commit, it's in the legacy compact anyway. Works, doesn't crash. Also, mind explaining how to attach commits to replies?

DonBruce64:
I don't now how to attach a commit to a reply.  But I can always just re-review the thing when I see the reply noted, like I do now.  Looks good to me sir!  Merged.

PR #2062, opened 2026-03-16
Fixed Explosions + external Damage completely blending a vehicle

SnailPerson:
check if damage is explosive, then only apply to core vehicle and skip dmg forwarding


DonBruce64 (changes requested):
Requesting changes based on prior change comments.

PR #2076, opened 2026-04-08
Added a custom explosion system with parameters from explosionOverhaul

dldev32:


DonBruce64:
Seems legit from an implementation standpoint.  I don't see how this would impact existing code badly.  It's just a new feature to play and tweak with that leaves old features intact.  Just make sure to update the Wiki with this please.

PR #2077, opened 2026-04-09
Removed Road Entry in car handbook

SnailPerson:
unfinished ass feature.
Perople keep asking about it and we gotta tell them they're all broken. 
This removes the roads entry to prevent that.


Wartich:
> So roads arent that broken really. Just folks don't use them much. What's broken about them you see?

They kinda are, breaking them freezes the server tps, they lag in general and kinda... Just don't have a proper purpose right now? They can't be used for railroads properly since trolleys and trains in iv would be able to drive over normal terrain. Can't really be used for traffic since there's no self driving and mts cars are a bit too heavy to be used as just traffic. And as roads for players they're just kind of meh because the placement is strange, there's no preview, breaking lags the server and singleplayer, so even with packs with roads people still mostly just build roads out of blocks. But many people have asked about roads on different IV discord servers, so since they're not really viable right now, maybe it makes sense to, at least temporarily remove them from the book until someone comes to refactor them into something more player friendly

DonBruce64:
A well-thought out and valid reply.  PR understood and approved.

PR #2078, opened 2026-04-09
Added damage to entities from collisions with vehicle collisions

dldev32:
Works if the collision type "entity" is specified.

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
@@ -33,6 +33,7 @@ import minecrafttransportsimulator.packets.instances.PacketPlayerChatMessage;
 import minecrafttransportsimulator.packets.instances.PacketVehicleServerMovement;
 import minecrafttransportsimulator.systems.ConfigSystem;
 import minecrafttransportsimulator.systems.LanguageSystem;
+import minecrafttransportsimulator.systems.LanguageSystem.LanguageEntry;
 
 /**
  * At the final basic vehicle level we add in the functionality for state-based movement.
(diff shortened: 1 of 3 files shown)


LauraDarkez:
how does this interfere with the "externalEntityDamage" function? because if you make it so all entity hitboxes deal damage to entities, wont that make it impossible to have vehicles where players can stand on top or inside them? and how does it effect riders?

dldev32:
> how does this interfere with the "externalEntityDamage" function? because if you make it so all entity hitboxes deal damage to entities, wont that make it impossible to have vehicles where players can stand on top or inside them? and how does it effect riders?

This only works when hitting at high-medium speeds. "externalEntityDamage" simply applies damage every tick, regardless of the conditions.

LauraDarkez:
but it doesnt apply damage to entities who are seated, right?

dldev32:
> but it doesnt apply damage to entities who are seated, right?

Absolutely correct.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleD_Moving.java
```diff
@@ -245,7 +246,7 @@ public void doPostUpdateLogic() {
             encompassingBox.heightRadius -= 1.0;
             for (IWrapperEntity entity : nearbyEntities) {
                 //Only move Vanilla entities not riding things.  We don't want to move other things as we handle our inter-entity movement in each class.
-                if (entity.getEntityRiding() == null && (!(entity instanceof IWrapperPlayer) || !((IWrapperPlayer) entity).isSpectator())) {
```
DonBruce64, line 248: You can't remove this code.  If the entity isn't a player it will crash trying to cast them to a player to check if they are a spectator.

DonBruce64 (changes requested):

PR #2086, opened 2026-04-25
Vehicle Dismount Safety Feature

SnailPerson:
After accidentally leaving my plane fro the second time and dying to it, I decided something must be done about that.

Basically a feature that will prompt the rider to press shift a second time to confirm dismount, when the vehicle is above 8ms fast, within the timeframe of a three second cooldown.

https://github.com/user-attachments/assets/1ab169c2-7b70-4c46-9fc0-9c9cf9a7e578


DonBruce64:
I see someone is making good use of that new overlay chat message function!

mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONConfigSettings.java
```diff
@@ -54,6 +54,7 @@ public static class ConfigGeneral {
         public JSONConfigEntry<Double> engineBiomeTempFactor = new JSONConfigEntry<>(1.0D, "Factor for how biome temp affects engine temp.  Higher values will make engines heat up quicker in hotter biomes.");
         public JSONConfigEntry<Double> rfToElectricityFactor = new JSONConfigEntry<>(0.02D, "Factor for converting RF to internal electicity for vehicles.  Default value is 1/100, but can be adjusted.");
         public JSONConfigEntry<Double> vehicleDeathDespawnTime = new JSONConfigEntry<>(0.0D, "Time (in seconds) between when vehicles reach 0 health and they de-spawn.  Normally 0, which means they never de-spawn.");
+        public JSONConfigEntry<Double> DismountSafteySpeed = new JSONConfigEntry<>(8.0D, "Speed, in m/s, above which vehicle dismounting requires pressing Shift twice within 3 seconds.  Set to 0 to always prompt, or a very high value to effectively keep classic instant dismounts.");
```
DonBruce64: This should go into the client config, not general.  Reason being the dismount setting is really something that each person should configure to their taste, not something that should be common to the pack or server.  Client config files are designed to not be included in modpacks just for this reason.

DonBruce64 (changes requested):

DonBruce64:
Also, I must say, I didn't know one could intercept shift key presses.  Possible use for walkable floors one day perhaps?

mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONConfigSettings.java
SnailPerson: fair point, will patch. Also good to see you again, its been a while

DonBruce64: One of these days I'll either need to fix some of these bugs, or ignore PRs till they get fixed. ;). But I'm not at either yet so PRs I review.

SnailPerson:
@dldev32 for your settings panel overhaul, the ms speed for dismount safety is now in client config. Its best to pull the settings again once all commits are merged

SnailPerson:
@DonBruce64 should be good?

DonBruce64:
@SnailPerson Ah, no.  Mis-understanding.  I was saying keep those functions as you have them, but move both of those configs to the Client config JSON file.  Not keep them in general.

That being said, if you don't see a reason for time being configurable I'm good with that.  Just put the speed in the client config or hard-code depending on how you see fit.  It's just them being in the general vs client config that has me comment.

SnailPerson:
@DonBruce64 yeah, all functions were kept as is, I just moved the x m/s for triggering the safety config from general to JSONClientConfig like you asked.

Or do you mean an additional setting for the three seconds? Or even having dismount safety in the first place?

Im a bit confuzzled haha..
As far as I know I didn't add a config for the cooldown time.

DonBruce64:
@SnailPerson you're right, I'm dumb.  All good!

PR #2087, opened 2026-04-27
New "guiTextObjects":[] Function + Collision Group Hover Variable Tweaks

SnailPerson:
PR adds two pack-authoring/runtime features aimed at cockpit/interactable UX without changing existing behavior:

1. `guiTextObjects`
2. box-specific collision hover variables

New Function: `guiTextObjects`
-  `rendering.guiTextObjects` section alongside existing `textObjects`.
- This is intended for screen-space text prompts tied to vehicles/parts, similar to existing HUD/overlay prompts, rather than world-space text rendered on the model.
- The implementation reuses the existing `JSONText` structure where possible, so pack authors can use the same text settings they already know:
  - `defaultText`
  - `fontName`
  - `scale`
  - `wrapWidth`
  - `renderPosition`
  - `color`
  - `variableName` / `variableFormat`
- Added GUI-specific support fields on `JSONText`:
  - `animations`
  - `fadeInTime`
  - `fadeOutTime`
- Rendering is client-side only and is done through the overlay GUI path, not the entity/world model renderer.
- Coordinates are screen-space, relative to screen center:
  - `+X` = right
  - `+Y` = down
  - `scale 1.0` = normal GUI font size
- Visibility/fading is animation-driven:
  - if animations resolve true, the text fades in
  - if they resolve false, the text fades out
  - default fade time is 7 ticks if not specified
- Variable-backed GUI text uses the same formatting path as normal text objects, so existing variable formatting behavior stays consistent.
- Added alpha-aware text rendering so GUI prompts can fade rather than pop.

All of this is used for fancy on-screen text, like instruments but for funny shti like this:

https://github.com/user-attachments/assets/aedd173a-5566-4916-a0ce-e36c3c54d05f

Visibility / player scoping
- `guiTextObjects` are rendered only on the local client overlay.
- Hover-driven prompts such as `player_cursor_hovered` and `collision_*_player_cursor_hovered` are therefore local to the player actually hovering the control.
- Added a GUI-hidden check so these prompts do not render when the player hides the HUD/GUI with F1.

## Aklo added box xsepcific collision hover variable
- Existing behavior remains unchanged:
  - `collision_<group>_player_cursor_hovered` still becomes true when *any* box in that collision group is hovered.
- Added optional per-box hover support:
  - `collision_<group>_<box>_player_cursor_hovered`
- `<box>` is 1-based and follows the order of the `collisions` array inside that collision group.
- The overlay hover tracking now sets/clears both:
  - the existing group-wide variable
  - the new exact-box variable
- Moving between boxes in the same group now correctly clears the previous box-specific variable and sets the new one.

This basically allows pack authors to attach prompts/animations to a specific collision box without breaking existing group-based setups. Or having to create an entirely different collision group for one hover text.

 also added `isGUIHidden()` support across all maintained interface modules so the F1/hidden-GUI behavior works on every supported version

- Existing `textObjects` behavior is unchanged.
- Existing `collision_<group>_...` variables are unchanged.
- The new features are additive:
  - packs that do not use `guiTextObjects` are unaffected
  - packs that do not use `collision_<group>_<box>_...` are unaffected

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
@@ -86,6 +86,10 @@ public abstract class AEntityD_Definable<JSONDefinition extends AJSONMultiModelP
      * Map containing text lines for saved text provided by this entity.
      **/
     public final Map<JSONText, String> text = new LinkedHashMap<>();
+    //GUI text objects keep their own animation and fade state as they are rendered independently from model text.
+    public final Map<JSONText, AnimationSwitchbox> guiTextSwitchboxes = new LinkedHashMap<>();
+    private final Map<JSONText, Double> guiTextAlphaValues = new HashMap<>();
+    private final Map<JSONText, Double> guiTextLastSampleTimes = new HashMap<>();
 
     /**
      * Map of computed variables.  These are computed using logic and need to be re-created on core entity makeup changes.
@@ -171,6 +175,7 @@ public abstract class AEntityD_Definable<JSONDefinition extends AJSONMultiModelP
     };
 
     public static final String REPAIRED_NAME = "repaired";
+    private static final int DEFAULT_GUI_TEXT_FADE_TICKS = 7;
 
     /**
      * Constructor for synced entities
@@ -326,6 +331,15 @@ public abstract class AEntityD_Definable<JSONDefinition extends AJSONMultiModelP
                     }
                 }
             }
+
+            if (definition.rendering.guiTextObjects != null) {
+                for (JSONText guiTextDef : definition.rendering.guiTextObjects) {
+                    if (guiTextDef.animations != null) {
+                        //GUI text only uses the switchbox result for visibility, but keeps the normal animation syntax.
+                        guiTextSwitchboxes.put(guiTextDef, new AnimationSwitchbox(this, guiTextDef.animations, null));
+                    }
+                }
+            }
         }
     }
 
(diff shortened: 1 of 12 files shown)


LauraDarkez:
this is a really nice feature

SnailPerson:
@DonBruce64 please check

DonBruce64:
I don't see anything inherently wrong with this from a programming standpoint.  The only comment I have, and I'm not going to make it required but rather a thought on something, is the fact that animations are packed into the text object file for GUI stuff, but are outside the file for non-GUI stuff.  Perhaps it would be better to have a separate JSON file for text animations on the GUI vs putting them into the TextObjects?  Though I can see that being a bit of an annoyance for PAs dealing with nested parentheses and is more something to do with object structure stuff vs a usability thing.

Food for thought. One thought, lemme know your thoughts.  I'll approve this as-is though if you feel it's better.

DonBruce64 (commented):
See comment.

SnailPerson:
@DonBruce64 they dont go into text objects. Its a separate section called guiTextObjects:[...]

Here an example:
```
"guiTextObjects": [	
  		{
    	"pos": [0, -24, 0],
    	"rot": [0, 0, 0],
    	"scale": 0.75,
    	"renderPosition": 0,
    	"defaultText": "Lights",
    	"color": "FFFFFF",
    	"fadeInTime": 4,
    	"fadeOutTime": 4,
    	"animations": [
      		{
        	  "animationType": "visibility",
        	  "variable": "collision_11_player_cursor_hovered",
        	  "clampMin": 1.0,
        	  "clampMax": 1.0
      		}
    	  ]
  		}
	],
```

Additionally, it has already been playtested on a private server with multiple members and no issues arose ;D

DonBruce64:
@SnailPerson  so I mean you put the bits in `mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONText.java`.  So even though you name it `guiTextObjects` it's the same as a `textObjects` on a vehicle.  Again though, if everyone is cool and PAs don't find it confusing, I is good to merge.  PAs good with it?

SnailPerson:
@DonBruce64 yea they cool with it, already did the wiki entry for it too ;D
additionally, I find it to be better separated anyway. Text objects are the physical ones, guiTOs are the guy ones, not present physically, so the separation makes sense :)

DonBruce64:
Le approved.

SnailPerson:
awesome!

PR #2089, opened 2026-05-03
Added selector for ammo and guns

dldev32:
The new selector introduces new capabilities for the mod's guns:
- Cycling through controller guns using keys 1-9;
- Changing ammo by pressing a key, unloading the rounds into available inventory (including gun-bound vehicle containers);
- Support for hand-held gun;
- The old gun indicator overlay has been removed.


LauraDarkez:
i found some issues with this pr
A: if you open the p menu, inventory or crate with a gun in hand or selected while in a vehicle, it breaks the text
B: this new ammo select does not work well on handheld guns when you have two ammo types and run out of the one you had selected. reload was blocked despite the secondary compatible ammo type being "selected"
C: this new ammo select does not allow for multi-type ammo loads on clipless weapons

LauraDarkez:
image
image
image

an example of broken text/gui. the panel also appears to break

DonBruce64:
I'm not opposed to this in principal, but two things I should note are that you don't want to use MC hotkeys as a keybind since clicks and such and changing what the player is holding can cause issues, such as using or clicking an item.  I still feel a single key you could press multiple times to index through to be a more fool-proof solution.

As to the changing ammo via a different key, that I get and should work fine with existing clip-type things as long as tested.  Will leave such testing to Laura and Conman and Cowboy who are more into guns and clips than I am.

dldev32:
> I'm not opposed to this in principal, but two things I should note are that you don't want to use MC hotkeys as a keybind since clicks and such and changing what the player is holding can cause issues, such as using or clicking an item. I still feel a single key you could press multiple times to index through to be a more fool-proof solution.
> 
> As to the changing ammo via a different key, that I get and should work fine with existing clip-type things as long as tested. Will leave such testing to Laura and Conman and Cowboy who are more into guns and clips than I am.

Unfortunately, I had to slightly modify the gun loading system because:
1) The IV lacks a unload mechanic for mounted guns;
2) The IV lacks a "round in the chamber" mechanic.

DonBruce64:
That's fair about the modifications.  And I get doing things in the code for that.  Not saying that's not allowed.  It's the interface that I'm more concerned about.  I've played mods where you use the numbers to select things and it usually messed up stuff when you're trying to click or fire with your gun and you select something else in your hand that opens a GUI or something.

The actual changes are welcome with that new GUI style especially the icons and such.  Much more professional than what I did.  So those I don't have comments on.  It's just the buttons mainly and having PAs's cross-check compat to make sure no bugs snuck in.

dldev32:
@DonBruce64 done.

dldev32:
@DonBruce64 maybe this time?

DonBruce64:
@dldev32 Git looks good.  @LauraDarkez does this no longer cause issues?

LauraDarkez:
as far as i know this is fixed, but i dont know for sure if i have gotten to test this branch's most recent adaptation

SnailPerson:
@DonBruce64

DonBruce64:
@SnailPerson  Did you test?  Laura said she didn't have a chance to test the latest thing.  Want to make sure we don't have another whoopsie like arcade mode stuffs.

Also, Git says branch conflicts, likely because this code touches the same code that was touched in another PR and it knows one of them can't work at the same time as the other.  Further making me  sussy about pulling it in, even if I could, and I can't, because it's got conflicts.

SnailPerson:
oh i See, sorry @DonBruce64, I dint see this comment here

PR #2094, opened 2026-05-12
Fixed issue that stopped mccore from compiling

SnailPerson:
Immediate compile failure in GUIOverlay.java (line 264), because InterfaceManager.clientInterface is typed as IInterfaceClient, and IInterfaceClient did not declare isGUIHidden().


DonBruce64:
Womp womp.  Wonder how that happened?

PR #2095, opened 2026-05-13
Fixed new explosion system

dldev32:
Here's what's been fixed:
* Fixed a crash when blastStrength radii were missing in XZ and Y;
* The explosion damage shape for blocks is now spherical, and raycast is only used to detect entities behind blocks;
* Blocks no longer break underwater;
* Added a new method for breaking blocks, which is much more performant.


mccore/src/main/java/minecrafttransportsimulator/mcinterface/AWrapperWorld.java
```diff
@@ -307,16 +307,26 @@ public abstract class AWrapperWorld extends EntityManager {
...
+    public abstract void destroyBlock(Point3D position, boolean spawnDrops);
+
+    /**
+     * Destroys the block at the position without running the normal block-break effects,
+     * including per-block sounds and break particles.  Used for bulk block changes such
+     * as explosions where per-block effects can overflow or stall clients.
+     * Wrappers should override this when their MC version exposes a quiet remove path.
+     */
+    public void destroyBlockQuietly(Point3D position, boolean spawnDrops) {
+        destroyBlock(position, spawnDrops);
```
DonBruce64: Given that all current interfaces have the ability to destroy blocks quietly, and that there is likely to be this behavior in all future versions of MC since it would make MC explosions that much more laggy, I propose this be an abstract method to match the rest of the wrapper code.  Interfaces can call the standard destroy method if needed, but one should be writing interfaces to challenge interface providers to provide a silent way to destroy blocks, not just auto-routing and not forcing at least some definement.

DonBruce64 (changes requested):

dldev32:
@DonBruce64 done.

PR #2096, opened 2026-05-14
Added "defaultZoom" to "Motorised"

SnailPerson:
added motorized.defaultZoom in JSONVehicle.java, 
applied that default in PartSeat.java, and changed AEntityB_Existing.java to always save/load zoomLevel so dev-mode imports keep the current zoom instead of resetting it. 
In JSON, "defaultZoom": X goes under motorized, 0 = the closest camera position.

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityB_Existing.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityB_Existing.java
@@ -115,7 +115,7 @@ public abstract class AEntityB_Existing extends AEntityA_Base {
             this.position = data.getPoint3d("position");
             this.orientation = new RotationMatrix().setToAngles(data.getPoint3d("angles"));
             this.motion = data.getPoint3d("motion");
-            this.zoomLevel = data.hasKey("zoomLevel") ? Math.max(0, data.getInteger("zoomLevel")) : 0;
+            this.zoomLevel = data.getInteger("zoomLevel");
             this.cameraIndex = data.getInteger("cameraIndex");
             this.radioData = data.getData("radio");
         } else {
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityB_Existing.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityB_Existing.java
@@ -115,7 +115,7 @@ public abstract class AEntityB_Existing extends AEntityA_Base {
             this.position = data.getPoint3d("position");
             this.orientation = new RotationMatrix().setToAngles(data.getPoint3d("angles"));
             this.motion = data.getPoint3d("motion");
-            this.zoomLevel = data.getInteger("zoomLevel");
+            this.zoomLevel = data.hasKey("zoomLevel") ? Math.max(0, data.getInteger("zoomLevel")) : 0;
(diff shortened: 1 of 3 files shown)


SnailPerson:
@DonBruce64

mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityB_Existing.java
```diff
@@ -115,7 +115,7 @@ public AEntityB_Existing(AWrapperWorld world, IWrapperNBT data) {
             this.position = data.getPoint3d("position");
             this.orientation = new RotationMatrix().setToAngles(data.getPoint3d("angles"));
             this.motion = data.getPoint3d("motion");
-            this.zoomLevel = data.getInteger("zoomLevel");
+            this.zoomLevel = data.hasKey("zoomLevel") ? Math.max(0, data.getInteger("zoomLevel")) : 0;
```
DonBruce64: Correct me if I'm wrong as it's been a while, but can you not just get the integer value as-is?  If it's not there it'll return 0 on that function, and why would this ever be negative?

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartSeat.java
```diff
@@ -47,6 +47,9 @@ public final class PartSeat extends APart {
 
     public PartSeat(AEntityF_Multipart<?> entityOn, IWrapperPlayer placingPlayer, JSONPartDefinition placementDefinition, ItemPartSeat item, IWrapperNBT data) {
         super(entityOn, placingPlayer, placementDefinition, item, data);
+        if (vehicleOn != null && (data == null || !data.hasKey("zoomLevel"))) {
+            this.zoomLevel = Math.max(0, vehicleOn.definition.motorized.defaultZoom);
```
DonBruce64: No need to math-max here either.  Default zoom should never be negative.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityB_Existing.java
SnailPerson: fair point, it was just a prevention in case it did somehow end up negative if a PA accidentally sets defaultZoom (new variable) of a vehicle into negative values
 
 Should I revert it to this.zoomLevel = data.getInteger("zoomLevel");?

DonBruce64: I would say so.  The assumption is folks don't do silly things, and if for some reason they wanted to make it negative, what's the harm there?

SnailPerson:
@DonBruce64 all good?

SnailPerson:
daddy @DonBruce64 please return

SnailPerson:
@DonBruce64 its been a week
please check in

SnailPerson:
@DonBruce64 please take 10 minutes to have a look here. We are waiting and it stalls further implementation because edited files start to overlap

SnailPerson:
@DonBruce64 ffs, ONE MONTH

SnailPerson:
@DonBruce64 
image

DonBruce64:
Somehow I wasn't getting emails for this one.  guess somehow I got un-subbed?

PR #2097, opened 2026-05-14
new config_x variables + 1 client config setting + render config tweaks + desc fix

SnailPerson:
Added variables:
`config_arcade` returns 1 when its active, like config_simplethrottle
`config_tutorial` returns 1 when its active
`config_autostartengines` returns 1 when its active

Added **client config**:
showTutorial - a simple 1 and 0 switch, used to trigger tutorial texts client side.
Goal is accessibility, along with the newly addd arcade mode. We are making IV optionally easier to understand for new users to hopefully grow its audienece and breathe new life in the low morale of the PA circle 

Transparent panel rendering option is now true by default, as many people complained about it being obstructive.
Also found a typo where 3p setting description was saying for 1st person despite being for 3rd.

https://github.com/user-attachments/assets/fa79af67-e0ee-4569-ba99-1ccebd0a103c


SnailPerson:
wiki been updated @DonBruce64

SnailPerson:
@DonBruce64 check an merge pwease <3

PR #2099, opened 2026-05-16
Experimental Config Parsing Fallback System

SnailPerson:
test test

Ive let my clanker loose onto the codebase with the goal of having an auotomated way that any missing or malformed configs are either added or saved with a fresh one being generated (without the old one getting lost)

I really did no quality checks here, this is purely experimental, but for what its worth, it works flawlessly for all config files

@DonBruce64 @liminal2

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/systems/ConfigSystem.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/systems/ConfigSystem.java
@@ -1,11 +1,19 @@
 package minecrafttransportsimulator.systems;
 
 import java.io.File;
+import java.io.IOException;
+import java.nio.file.AtomicMoveNotSupportedException;
 import java.nio.file.Files;
+import java.nio.file.StandardCopyOption;
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import minecrafttransportsimulator.items.components.AItemPack;
 import minecrafttransportsimulator.items.components.AItemSubTyped;
@@ -17,6 +25,7 @@ import minecrafttransportsimulator.jsondefs.JSONConfigCraftingOverrides.JSONCraf
 import minecrafttransportsimulator.jsondefs.JSONConfigExternalDamageOverrides;
 import minecrafttransportsimulator.jsondefs.JSONConfigSettings;
 import minecrafttransportsimulator.jsondefs.JSONSubDefinition;
+import minecrafttransportsimulator.mcinterface.IWrapperPlayer;
 import minecrafttransportsimulator.mcinterface.InterfaceManager;
 import minecrafttransportsimulator.packloading.JSONParser;
 import minecrafttransportsimulator.packloading.PackParser;
@@ -39,6 +48,8 @@ public final class ConfigSystem {
     public static JSONConfigClient client;
     public static JSONConfigExternalDamageOverrides externalDamageOverrides;
     private static File configDirectory;
+    private static final List<String> pendingConfigWarnings = new ArrayList<>();
+    private static final Set<String> blockedConfigSavePaths = new HashSet<>();
 
     /**
      * Called to load the config objects from the files in the passed-in folder.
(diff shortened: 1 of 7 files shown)


SnailPerson:
@DonBruce64 
image
image

SnailPerson:
[Immersive Vehicles-1.21.1-25.0.0.jar.zip](https://github.com/user-attachments/files/27863364/Immersive.Vehicles-1.21.1-25.0.0.jar.zip)

@liminal2 please try this beta :)

mccore/src/main/java/minecrafttransportsimulator/systems/ConfigSystem.java
```diff
@@ -39,46 +48,29 @@ public final class ConfigSystem {
...
-                    InterfaceManager.coreInterface.logError(e.getMessage());
-                }
-            }
-            if (client == null) {
-                client = new JSONConfigClient();
-            }
+            client = loadConfig(clientFile, JSONConfigClient.class, new JSONConfigClient(), true);
             //Check to make sure we have the right keyset.  If not, reset the binds.
-            if (client.controls.keysetID != InterfaceManager.inputInterface.getKeysetID()) {
+            if (client.controls == null || client.controls.keysetID != InterfaceManager.inputInterface.getKeysetID()) {
```
DonBruce64, line 73: This for sure fixes fresh configs automatically generating an error on first launch.

mccore/src/main/java/minecrafttransportsimulator/systems/ConfigSystem.java
```diff
@@ -102,141 +94,282 @@ public static void initOverrides() {
...
-                        Map<String, JSONCraftingOverride> packOverrides = overridesObject.overrides.get(packItem.definition.packID);
-                        if (packOverrides != null) {
-                            JSONCraftingOverride override = packOverrides.get(packItem.definition.systemName);
-                            if (override != null) {
-                                if (override.commonMaterialLists != null) {
-                                    packItem.definition.general.materialLists = override.commonMaterialLists;
+            try {
+                JSONConfigCraftingOverrides overridesObject = loadConfig(craftingFile, JSONConfigCraftingOverrides.class, createEmptyCraftingOverrides(), true);
+                if (overridesObject.overrides == null) {
+                    overridesObject.overrides = new LinkedHashMap<>();
```
DonBruce64, line 114: What's this do?  Makes a new overrides object that's empty?  Wouldn't you just want to re-create the file like before?

DonBruce64 (commented):
Looks okay, though would like a small spiel on how this works prior to merging.  If it does work, which it would be easy to test, then I'm good merging.  Just would like to know code flow first.

mccore/src/main/java/minecrafttransportsimulator/systems/ConfigSystem.java
SnailPerson: That line creates the empty map that the next loop fills. It is not meant to stay empty.
So new LinkedHashMap<>() is just initializing the container. Without it, overridesObject.overrides.computeIfAbsent(...) would crash with a null pointer. The file is still recreated “like before” because the loop populates that map before writing it to disk.

basically
If mtscraftingoverrides.json exists and parses, loadConfig returns the user’s parsed override file.
If it is malformed, loadConfig backs it up, creates a fresh replacement from the supplied default object, and returns that default.
If it is missing, loadConfig(..., true) creates a fresh replacement from the supplied default object and returns that default.
createEmptyCraftingOverrides() is intentionally empty: no overrides apply unless the user/modpack put overrides there.
The null guard handles a valid JSON file that omits "overrides" or has it as null, so the later overridesObject.overrides.get(...) does not crash.

You would not want createDefaultCraftingOverrides() there unless you want missing or malformed mtscraftingoverrides.json to be replaced with a fully populated dump of every pack item recipe. That is what the old generateOverrideConfigs path does. For normal startup, that would be more invasive because it creates a huge template file and can look like IV is applying/generated lots of override data when no override was requested.

So the split is intentional:

generateOverrideConfigs = true: recreate the full template using createDefaultCraftingOverrides().
normal load/missing/malformed recovery: create an empty, valid override file so the game can continue without applying overrides.

@DonBruce64 #

DonBruce64:
@SnailPerson if you're sure this is all gucci and doesn't break things with various config issues, I'm good to merge.  Lemme know.

SnailPerson:
@DonBruce64 well I suppose? It is a hefty diff though and im only sure if you looked into the code and decided it seems waterproof.

DonBruce64:
@SnailPerson It doesn't look problematic, and if it's tested, I'm good with it.  I can't see too much different I would have done.

PR #2100, opened 2026-05-17
Updated the culling Warning and added a client config to disable it

SnailPerson:
added a client config option to disable reoccurring entity culling messages
Updated warn message slightly. 
Beforr:
IV HAS DETECTED THAT ENTITY CULLING MOD IS PRESENT.  THIS MOD CULLS ALL IV VEHICLES UNLESS \"mts:builder_existing\", \"mts:builder_rendering\", AND \"mts:builder_seat\" ARE ADDED TO THE WHITELIST.

Its now: 

ENTITY CULLING MOD IS PRESENT.  WHITELIST \"mts:builder_existing\", \"mts:builder_rendering\", AND \"mts:builder_seat\". IN CONFIG FILE OR VEHICLES MAY BE CULLED. (You can turn off this message in IV's client config menu.)

also no idea why diffs are so lagre, I suspect GitHub is doing some formatting funkies


DonBruce64:
Big diffs are probably due to spaces or tabs or similar formatting differences.  Especially if your code editor does them differently than another user.  I noticed DL Dev had that happen too, likely for the same reason.

PR #2103, opened 2026-05-19
Experimental Sound Code Optimisations

SnailPerson:
WARNING: @DonBruce64 
THERE ARE multiple PRs editing entityD_definable by now and I hope implementation goes without overwriting another PRs work! Please make sure no changes are lost please.
Same with config client etc.

## Summary

Refactors entity sound updates to explicitly separate tick-only sounds from partial-tick sounds, and prevents partial-tick sound logic from running more than once per rendered frame.

## Problem

Before this change, all JSON-defined sounds were stored in a single list and filtered inside `updateSounds(float partialTicks)`.

That had two issues:

1. Tick-only sounds were still paying render-path overhead just to be skipped.
2. Sounds with `canPlayOnPartialTicks: true` could be evaluated more than once per frame because the render path runs in multiple passes.

Entity sounds are updated from both the client tick path and the render path:
- client tick: AEntityB_Existing.java:163
- render path: AEntityC_Renderable.java:113

The render path itself can run in both solid and blended passes, which made partial-tick sounds do duplicate work in the same frame.

## What Changed

In AEntityD_Definable.java:99

- Replaced the single `allSoundDefs` list with:
  - `tickSoundDefs`
  - `partialTickSoundDefs`
- Classified each `JSONSound` once during entity setup based on `soundDef.canPlayOnPartialTicks` instead of re-checking that flag in the hot update loop.
- Updated `updateSounds(float partialTicks)` to select the correct sound list up front rather than iterating every sound definition on every call.
- Added a frame guard using:
  - `lastTickPartialSoundsUpdated`
  - `lastPartialTickSoundsUpdated`

This ensures partial-tick sounds are only processed once per rendered frame, even if the entity is rendered multiple times in separate passes.

## Why

This reduces avoidable work in the entity sound hot path while preserving the existing pack-facing behavior.

Benefits:
- tick-only sounds no longer incur unnecessary render-side iteration
- partial-tick sounds no longer do duplicate evaluation in multi-pass rendering
- the sound update flow is easier to reason about and maintain

## Behavior / Compatibility

This change does not modify the JSON sound format and does not require any content pack changes.

Behavior remains the same for pack authors:
- normal sounds still update on ticks
- `canPlayOnPartialTicks` sounds still update during rendering

The only behavioral correction is that partial-tick sounds are no longer evaluated twice in the same frame due to multiple render passes.

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityD_Definable.java
@@ -96,7 +96,8 @@ public abstract class AEntityD_Definable<JSONDefinition extends AJSONMultiModelP
      **/
     protected final Map<String, ComputedVariable> computedVariables = new HashMap<>();
 
-    private final List<JSONSound> allSoundDefs = new ArrayList<>();
+    private final List<JSONSound> tickSoundDefs = new ArrayList<>();
+    private final List<JSONSound> partialTickSoundDefs = new ArrayList<>();
     private final Map<JSONSound, AnimationSwitchbox> soundActiveSwitchboxes = new HashMap<>();
     private final Set<JSONSound> soundDefFalseLastCheck = new HashSet<>();
     private final Map<JSONSound, SoundSwitchbox> soundVolumeSwitchboxes = new HashMap<>();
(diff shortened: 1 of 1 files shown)


DonBruce64:
@SnailPerson Did this not work?  I totally get the idea behind it and support it.

PR #2105, opened 2026-05-19
Added player camera rotation bounds when using custom cameras on guns

dldev32:


mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
```diff
@@ -954,9 +957,28 @@ private void handleControl(IWrapperEntity controller) {
                     lastControllerSeat.riderRelativeOrientation.angles.x -= (internalOrientation.angles.x - prevInternalOrientation.angles.x);
                 }
             }
+            if (shouldLockControllerToActiveCamera()) {
```
DonBruce64: Don't sub-routine things only used in one spot unless it's a big section of code that needs to be sub-routined for clarity of code flow.  Boolean conditions should generally not be sub-routined since it just generated extra code calls.  I'm okay with the locking logic being sub-routined if you want it to be.

DonBruce64 (changes requested):
I'm assuming this is to force the camera to follow the gun orientation?  Doesn't it already do this?

dldev32:
> I'm assuming this is to force the camera to follow the gun orientation? Doesn't it already do this?

If we're talking about custom cameras, then yes. However, the player's camera can rotate freely outside the custom camera's rotation angles, which are limited by the variable values. So, I've corrected this oversight, and the player's camera can now sync with the custom camera.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/PartGun.java
dldev32: Done.

DonBruce64:
@LauraDarkez this true and this PR fix this?

LauraDarkez:
i havent gotten a jar to test. why are you pinging me?

LauraDarkez:
ive confirmed this seems to work as intended

PR #2107, opened 2026-05-22
Oh.. ravish me daddy

SnailPerson:
changed the C hotkey form toggling mouse yoke to toggling the new arcade mode.
Classic mouse yoke may now only be activated from the p menu.


SnailPerson:
@DonBruce64 hope everything is done correctly here, old mouse yoke is no longer tied to the C, arcade mode is instead
Please check

mccore/src/main/java/minecrafttransportsimulator/systems/LanguageSystem.java
```diff
@@ -369,14 +369,14 @@ public static class JSONLanguageFile {
...
+    public static final LanguageEntry INTERACT_VEHICLE_LOCKED = new LanguageEntry("interact.vehicle.locked", "This vehicle is locked!");
+    public static final LanguageEntry INTERACT_VEHICLE_NEEDKEY = new LanguageEntry("interact.vehicle.needkey", "You must have the key of this vehicle to start it!");
+    public static final LanguageEntry INTERACT_VEHICLE_NOSPACE = new LanguageEntry("interact.vehicle.nospace", "There is not enough space to place down this vehicle here.");
+    public static final LanguageEntry INTERACT_VEHICLE_JUMPERPACK = new LanguageEntry("interact.vehicle.jumperpack", "Charged vehicle battery to maximum.");
+    public static final LanguageEntry INTERACT_VEHICLE_NOFUEL = new LanguageEntry("interact.vehicle.nofuel", "There is no fuel in this vehicle.  The engine cannot start!");
+    public static final LanguageEntry INTERACT_VEHICLE_DISMOUNTCONFIRM = new LanguageEntry("interact.vehicle.dismountconfirm", "Press Shift again to dismount");
+    public static final LanguageEntry INTERACT_MOUSEYOKE_ENABLED = new LanguageEntry("interact.mouseyoke.enabled", "Arcade Mode enabled");
```
DonBruce64: Still says mouseyoke in the entry JSON section.

mccore/src/main/java/minecrafttransportsimulator/systems/LanguageSystem.java
```diff
@@ -369,14 +369,14 @@ public static class JSONLanguageFile {
...
+    public static final LanguageEntry INTERACT_MOUSEYOKE_DISABLED = new LanguageEntry("interact.mouseyoke.disabled", "Arcade Mode disabled");
```
DonBruce64: Still says mouseyoke in the entry JSON section.

DonBruce64 (changes requested):
In addition to these two comments, is there a reason you'd want to turn off and on either arcade mode or mouse yoke via a keybind?  Why not just have them both in the config menu?  I would think most players would just have a preference and stick with it, like how they do window rendering and such.

LauraDarkez:
having the keybind to disable is nice when you need to go from controling the plane to interacting with the cockpit/freely looking around. it is also nice to be able to switch to/from manual keybind control quickly when transitioning from flight to landing/takeoff to flight

SnailPerson:
@DonBruce64 daddy reevaluate please

Also hotjey, as Laura stated, is important because opening p and clicking thrice is hella annoying while flying a plane and trying to land it etc. Its much quicker and easier

LauraDarkez:
snail, what is with your remarks towards don? i cant tell if its a kink or a meme at this point

DonBruce64:
@SnailPerson Totally get not wanting to open a menu for that.  But do we need to keep the keybind for mouseYoke?  What if you want to switch in and out of looking vs controlling for that?  Isn't mouseYoke just diet arcade mode?  Or does arcade mode supplant mouseyoke?

SnailPerson:
@DonBruce64 mouse yoke is mouse yoke, arcade mode is better mouse yoke with more functionality. The hotkey for mouse yoke has been moved over to arcade since arcade is better, but in cases where classic mouse yoke is useful its still is in the p menu. Just not hotkeyable anymore because it has been "replaced" by a "better mouse yoke system".

The hotkey is important because for example, flying helis is not as easy with arcade, so toggling it quickly makes sense. Or wanting to do some keyboard manoeuvres. Or wanting to turn it off quickly to look around without affecting the planes direction and then reenabling it to continue piloting
In any case,i have play tested with a number of PAs on a private server on survival and it's genuinely a good change.
Nobody wants to press p, then click, then click again. By that time, your plane is already in a mountain.

DonBruce64:
@SnailPerson Totally get that.  My concern here was that I saw the mouseYoke hotkey go away and get replaced by an arcade mode hotKey.  Seeing that, I was concerned that people using mouseYoke would end up in just the situation you describe.  If someone wants to use mouseYoke, but not Arcade mode, it sounds like they can enable it in the P menu.  But then they can't dynamically turn it off anymore with the hotkey?  Or does the hotkey turn mouseYoke and arcadeMode both on and off depending on what's in the P-key menu?

I would assume that the system would have a single key for enabling arcadeMode OR mouseYoke depending on what's selected in the P-menu.  Changing P-menu items from one to the other with one active would make the other active, or if neither were active would make the hotkey turn the requested one on or off.

SnailPerson:
@DonBruce64 arcade mode is basically mouse yoke but better. That's why it gets the hotkey privilege

DonBruce64:
@SnailPerson Right, but by doing this you have removed the mouseYoke hotkey for users who might want it.  They will have their planes fly into things just the way you described.  Either remove mouseYoke entirely so not to confuse players, or keep both bound to the same hotkey and just switch which is active in the P-key config menu.

SnailPerson:
# 🌚

SnailPerson:
@DonBruce64 please just merge this, its just a hotkey switch, the old yoke is being kept for situations in which it is actually still useful, but in most cases, arcade is superior, hence it being the default for the hotkey. As previously mentioned, this has been through play testing and approved as a superior method.
There's no proper argument against making the switch here. People can still play with the old yoke perfectly fine if they so desire, so were not taking away anything.

DonBruce64:
@SnailPerson What you're taking away is their ability to turn on and off yoke with a hotkey.  It's why I'd simply like the hotkey to toggle either arcade or yoke depending on which is active.  It's not hard to just re-name the hotkey to something generic, change the tooltip to say it toggles whichever is set via the P-key menu, and have the press action just check an if statement and do either one or the other when pressed.

It's not hard to add that, hence why I'm asking for those who still want mouse yoke. MTS is grounded on not removing old features or functionality unless they are game-breaking (like double-sided rendering making things white).  But since we can keep that option for players I would like us to.

SnailPerson:
@DonBruce64 again, mouse yoke was a placeholder for the arcade mode anyway. Keeping it is LC.
Please just merge it, it's not that deep of a topic. 

Beides, how would that look. If a player has mouse yoke and arcade enabled at the same time, triggered by the same hotkey, thats just chaos.

Old system got replaced by new system that does the same thing but better. That's all

DonBruce64:
I still disagree, but I'll relent since you're running the ship now and know the pulse better than me.  Merged!

PR #2109, opened 2026-05-26
Added help button to panels

SnailPerson:
This is part of the effort to make IV more accesible.
Helper text os one of the steps to make it easier for players to get into the mod, apart from updated guides, hud warns and tutorial segments.

https://github.com/user-attachments/assets/54224164-a3a8-4b87-b6e5-1f8f71a4d112

@DonBruce64


SnailPerson:
@DonBruce64 please review latest commits

DonBruce64:
Can merge as-is, but thoughts about it being a config in the P-menu?  Really I would assume that's where the majority of configurable things would belong.  Having it be in the panel by default requires it to be in all panels for all packs and that could lead to issues in vehicles where the PA wants either an invisible panel or fully immersible one without a help button.

SnailPerson:
@DonBruce64 this is more of a global effort. Im providing all PAs with snippets that use this helper button. 
IV is too complicated for the average user. This means that many people tend to be scared off. Any initiative to be more user friendly is a productive one. 

Applied methods are already showing in our numbers, IV has never done better, just recently we had peaks of 14-18k downloads per day. 
IV broke 7.200.000 downloads yesterday.

This is modrinth.
image

DonBruce64:
@SnailPerson Not saying that this isn't a good idea.  On the contrary, I do think it's a good idea.  The thing is I question making it a button on the panel.  Most overlays and tooltips and such are controlled from the config screen, so I would assume this one would be as well.  Plus, by doing it this way you have to hide the help on every vehicle you make since it'll be shown by default, and if another user in a multiplayer setting hides help, it won't be availabe on that vehicle to a new user who may want that help.  Hence me saying this should be player-centric not vehicle-centric for where the config is stored.

SnailPerson:
those are good points.
Config option is reasonable

PR #2112, opened 2026-05-30
Some bullet rework

dldev32:
Here are the changes:
* New bullet types have been added (some of which are needed to simplify obtaining bullet types);
* A raycast-based fragmentation system has been added. Fragments have a cone angle, quantity, damage, and range. Fragments are created only when armor is penetrated, damaging internal components and entities located within;
* Bullet hits on entities now occur at the current hit position, rather than at the feet of the hitbox;
* HEAT projectiles now stop upon collision and no longer pass through the entire vehicle. For behind-the-armor effects, it's sufficient to add a highly targeted set of high-damage fragments.

_This PR has numerous rewritten files to correct formatting issues._


mcinterfaceforge1122/src/main/java/mcinterface1122/InterfaceClient.java
DonBruce64, line 1: Something is wrong with your saving of these files.  It shows all lines are changed?

.gitattributes
DonBruce64, line 1: This... shouldn't ever change?

mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityF_Multipart.java
```diff
@@ -371,7 +378,96 @@ public EntityBullet.HitType attackProjectile(Damage damage, EntityBullet bullet,
                         } else {
                             EntityBullet.performGenericHitLogic(bullet.gun, bullet.bulletNumber, hitEntry.position, hitEntry.side, HitType.ARMOR);
                         }
-                        bullet.displayDebugMessage("HIT TOO MUCH ARMOR.  MAX PEN: " + (int) penetrationPotential);
+                        if (bulletIsHeat) {
```
DonBruce64, line 381: Noting this for review from another PA.

DonBruce64 (changes requested):

DonBruce64:
@LauraDarkez and @SnailPerson y'all probably wanna review this with bullets.  It's fairly big in terms of what's changed and probably needs to be tested.

SnailPerson:
@DonBruce64 this is a pretty old pr, and dldlv didn't give much context nor any betas to actually test this.
I wouldn't merge anything without properly testing it as you said, so best hold off for now.

DonBruce64:
@SnailPerson  Really all this does is changes how bullet pen operates on vehicles when penning multiple boxes.  Should be simple enough to test existing behavior.  If that stays functional, then I don't see why the playground can't be added for optional pen stuff.  But yes, test first before merging since I can't rightfully review this.

PR #2114, opened 2026-05-31
Added hitboxOffset parameter to part slot to move the position of the part's hitbox

dldev32:
This allows to move the hitbox of a part while keeping its model in its original position.

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
@@ -89,6 +89,7 @@ public abstract class APart extends AEntityF_Multipart<JSONPart> {
     public final RotationMatrix zeroReferenceOrientation;
     public final RotationMatrix prevZeroReferenceOrientation;
     public final Point3D externalAnglesRotated = new Point3D();
+    private final Point3D scaledHitboxOffset = new Point3D();
     private final AnimationSwitchbox placementActiveSwitchbox;
     private final AnimationSwitchbox internalActiveSwitchbox;
     private final AnimationSwitchbox placementMovementSwitchbox;
@@ -96,6 +97,9 @@ public abstract class APart extends AEntityF_Multipart<JSONPart> {
 
     public APart(AEntityF_Multipart<?> entityOn, IWrapperPlayer placingPlayer, JSONPartDefinition placementDefinition, AItemPart item, IWrapperNBT data) {
         super(entityOn.world, placingPlayer, item, data);
+        this.placementDefinition = placementDefinition;
+        this.placementSlot = entityOn.definition.parts.indexOf(placementDefinition);
+
         //Init locals.
         this.localOffset = placementDefinition.pos.copy();
         this.localOrientation = new RotationMatrix();
@@ -126,6 +130,7 @@ public abstract class APart extends AEntityF_Multipart<JSONPart> {
             scale.multiply(placementDefinition.partScale);
         }
         prevScale.set(scale);
+        updatePartBoundingBoxCenter();
 
         //Set switchbox animation properties.
         //Placement movement needs to take into account applyAfter animations.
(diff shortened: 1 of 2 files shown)


SnailPerson:
what use does this have
why not just move it with an animation or simply adjust the original position..-

dldev32:
> what use does this have why not just move it with an animation or simply adjust the original position..-

Because it's much cheaper for the CPU, and doing it through animations is a bit of a hassle, because essentially you're moving the model, and you need to somehow sync collisions with that model.

SnailPerson:
@dldev32 again, the exact use of this PR is unclear.
As i undersand this is an offset that offsets a hotboxes coordinates. 
What use does this have when you can simply edit the initial coordinates. Say I have a box at 1 1 1, and I put offset to 1 2 1, why dont I just set 1 2 1 form the very beginning?
Unless there's an aspect you did not mention?

dldev32:
> @dldev32 again, the exact use of this PR is unclear. As i undersand this is an offset that offsets a hotboxes coordinates. What use does this have when you can simply edit the initial coordinates. Say I have a box at 1 1 1, and I put offset to 1 2 1, why dont I just set 1 2 1 form the very beginning? Unless there's an aspect you did not mention?

I'm just too lazy to move the model in Blender, so I added this parameter to move the hitbox of a part, not the entire model. For example, you have a tank turret model, but its hitbox is in the center, and it somehow needs to be hidden down.

DonBruce64:
@SnailPerson Good call on this.  I remember we had thoughts about this before and settled on position standardization long ago.

@dldev32 the reason for the standardization and not allowing hitbox/part offsets was due to packs having different parts with different 0 points.  That caused issues with seats/crates not being pack-compatible.  I also forsee the same happening if people are changing hitboxes and not the models.  It could lead to un-intended issues.  While I appreciate the effort into this PR, I can't rightfully accept it since it leads to possible incompatibilities given past experiences.

dldev32:
> @SnailPerson Good call on this. I remember we had thoughts about this before and settled on position standardization long ago.
> 
> @dldev32 the reason for the standardization and not allowing hitbox/part offsets was due to packs having different parts with different 0 points. That caused issues with seats/crates not being pack-compatible. I also forsee the same happening if people are changing hitboxes and not the models. It could lead to un-intended issues. While I appreciate the effort into this PR, I can't rightfully accept it since it leads to possible incompatibilities given past experiences.

I don't see the connection between the incompatibility and the new parameter. If you don't want it, don't use it. What's the problem? I don't understand that.

I'll add: even if this innovation seems like nonsense to you, then offer me an alternative solution, but not the one that's currently being used and makes you spend a whole hour on only one model of vehicle, when it could be done thanks to the new parameter in literally a minute.

DonBruce64:
So, you make a fair point.  Lemme re-review this. I can see some use-cases provided one at least aligns their part points to where they need to be, even if the model is asymmetric.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
```diff
@@ -386,11 +377,30 @@ public boolean shouldSavePosition() {
...
+        return false;
+    }
+
     @Override
     public boolean canBeClicked() {
         return entityOn.isVariableListTrue(placementDefinition.interactableVariables) && entityOn.canBeClicked();
     }
 
+    private void updatePartBoundingBoxCenter() {
+        boundingBox.globalCenter.set(position);
```
DonBruce64, line 391: I'm assuming this doesn't cause issues by only updating the globalCenter and not setting the localCenter in the constructor?

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/entities/instances/APart.java
dldev32: Yeah, that's true.

DonBruce64:
@dldev32 merged.  please update wiki if you haven't already.

SnailPerson:
@dldev32 did you update the wiki

dldev32:
> @dldev32 did you update the wiki

Of course.

PR #2116, opened 2026-06-04
Gun Tweaks for the Peaks of Freaks

mklave4d6v8:
Adds a few parameters to guns and bullets

Bullet parameters:
damageFalloff - How much damage this bullet loses during its travel. If not defined, this bullet should always deal full damage.
damageFalloffTime - How long until this bullet's damage is reduced to damageFalloff's multiplier. If not defined, this parameter defaults to despawnTime.
damageFalloffDelay - How long until this bullet's damage starts to reduce.
trajectoryCurveX - How much velocity, each tick, should be applied in a random X direction.  Used to make bullets curve off their trajectory mid-flight. (These are similar in function to bulletSpreadFactor, but are NOT the same as it!)
trajectoryCurveY - How much velocity, each tick, should be applied in a random X direction.  Used to make bullets curve off their trajectory mid-flight.
trajectoryCurveDelay - How long, in ticks, the bullet should take before applying trajectory curves. Used to keep bullets straight for a set amount of time.

bulletDamageFactor - How much to multiply this bullet's damage. 0 is the bullet's defined amount, -1 will disable bullet damage, and 1 will deal double damage.

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/rendering/RenderText.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/rendering/RenderText.java
@@ -430,6 +430,11 @@ public class RenderText {
                 wrapWidth = 0;
             }
 
+            //Convert the scaled pixel width into the unscaled font coordinate space used
+            //to lay out glyphs.  Do this before the world-space 1/16 scale conversion:
+            //wrapWidth is defined in pixels for both GUI and world text.
+            wrapWidth /= scale;
+
             //Reduce scale by 16 if we're not using pixel coords.
             //Entity rendering systems calling this function feeding params will scale in block coords.
             //We need to be be in pixel coords for all the rendering operations here.
@@ -448,9 +453,6 @@ public class RenderText {
             } else if (alignment.equals(TextAlignment.RIGHT_ALIGNED)) {
                 alignmentOffset = -stringWidth;
             }
-
-            //Adjust wrapWidth to account for total scale.
-            wrapWidth /= scale;
 
             //The following rendering setup code operates on the assumption of a 1.0 scale.
             //This function sets up the font using the paramters defined above, and sets all UV points.


dldev32:
I don't think it's a good idea to simulate improved bullet ballistics using these methods. The current implementation already assumes that armor penetration drops as bullet velocity decreases. Why not simply extend the code and add the same for damage? Just like a curved trajectory, a bullet doesn't behave this way in flight if you set a predictable motion for it. Bullet lateral movement is usually affected by factors like wind, temperature, and other weather conditions, but this isn't implemented in the IV, even though it would be much better than simply a fixed spread/trajectory value.

mklave4d6v8:
1. Because I'm not aiming for ballistic simulation at all - I'm treating our bullet system like a conventional game would. On top of that, we do already reduce damage as the bullet loses velocity https://github.com/DonBruce64/MinecraftTransportSimulator/pull/2116/changes#diff-5e3c98a79b5c7255cf93e6314db8a52dac6af790fef8ef1e7b1c9cd163e4b6bdR349 https://github.com/boot2big/MinecraftTransportSimulator-CBfork/blob/d6b595af48cc67d5a2ce9c5023e8b96c66b739b3/mccore/src/main/java/minecrafttransportsimulator/entities/components/AEntityF_Multipart.java#L362 BUT this is not what I wanted. In addition, there's no easy way to do damage rampup for more arcadey style guns (which, if I recall, may have even been patched out anyways!) This ensures full control of how much *base damage* the bullet is supposed to do, and when; don't forget that this applies *before* the aforementioned velocity calculations.
2. You're right, we don't have atmospheric conditions. I don't want those! I want a way for my bullet to travel in a straight line for a certain amount of time, *then* have it deviate *later*. This way I can have, say, a shotgun that has tight spread in close quarters, manageable spread at the mid-range, before finally becoming useless at long-range. (As opposed to our current system where... you either get a sniper-shotty, or something useless outside of melee range.)

I'd be willing to work with you on an alternate code snippet for damage falloff, so long as it can handle negative falloff (rampup.) Bullet curve is non-negotiable, I've never truly been satisfied with how bullet spread plays out in MTS. I don't mind the initial spread itself, but I dislike losing any amount of spread control once the bullet leaves the gun... it doesn't have to _be realistic_, it just needs to _feel immersive_

LauraDarkez:
honestly, im agreeing with cb here, theres no reason this method cant be implemented for immersion and function over realism.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
```diff
@@ -187,6 +196,27 @@ public void update() {
                 motion.y -= definition.bullet.gravitationalVelocity;
             }
 
+            //Add spread forces, once we've counted down our spread timer.
+            if (ticksExisted > definition.bullet.trajectoryCurveDelay) {
+                motion.x -= trajectoryCurvatureVelocityX;
+                motion.y -= trajectoryCurvatureVelocityY;
```
mklave4d6v8: this moron didn't use her brain and applied cardinal motion relative to the world!! 

https://tenor.com/view/scream-laughing-brown-cat-lmao-rofl-gif-14057163617344906417

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
```diff
@@ -100,7 +106,10 @@ public EntityBullet(Point3D position, Point3D motion, RotationMatrix orientation
             velocityToAddEachTick = 0;
             motionToAddEachTick = null;
         }
+        this.trajectoryCurvatureVelocityX = Math.random() >= 0.5 ? definition.bullet.trajectoryCurveX : definition.bullet.trajectoryCurveX * -1;
```
mklave4d6v8: still too straight. it needs to be as gay and wiggly as the most peak identities of all time

mklave4d6v8 (commented):
i'm stupid and need to fix this

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
```diff
@@ -101,6 +108,7 @@ public EntityBullet(Point3D position, Point3D motion, RotationMatrix orientation
             motionToAddEachTick = null;
         }
         this.despawnTime = definition.bullet.despawnTime != 0 ? definition.bullet.despawnTime : 200;
+        this.damageFalloffTimer = definition.bullet.damageFalloffTime != 0 ? definition.bullet.damageFalloffTime : this.despawnTime;
```
DonBruce64: Should be damageFalloffTime.  Not a timer, this is a value.  Also helps to make it match JSON.

mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONBullet.java
```diff
@@ -44,6 +44,15 @@ public static class Bullet {
         @JSONDescription("How much damage this bullet does.  Is set to 1/5 the diameter if left out.  Note that 'water' type bullets don't damage things, no matter this value.")
         public float damage;
 
+        @JSONDescription("How much damage this bullet loses during its travel. If not defined, this bullet should always deal full damage.")
```
DonBruce64: Perhaps tweak this to mention that velocity plays a part in this "full damage" formula.  Or do you mean for entities?  Cause I thought we reduced damage for slower bullets for them and vehicles, so it would be important to note here in the JSON so PA's don't get confused why their math isn't mathing or they have damage reduction when this isn't set.

mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONPart.java
```diff
@@ -451,6 +451,9 @@ public static class JSONPartGun {
         @JSONDescription("How much spread the bullet will have when fired.  0 is no spread, higher values have higher spread.")
         public float bulletSpreadFactor;
 
+        @JSONDescription("How much to multiply this bullet's damage. 0 is the bullet's defined amount, -1 will disable bullet damage, and 1 will deal double damage.")
```
DonBruce64: I would LC this to a value of 1.  A factor should be from 0->1+, having 0 be standard and negative lower is confusing.  Also then -1 plays heck for code, one should just be able to set it to 0 and have it be no damage.  Though I suppose that would make LCs bring it back up to 1.  You thinking of having guns that by default don't deal any damage?  That's the only issue I could see with setting it this way.  But math-wise having it be 0 does standard doesn't make intrinsic sense, even if you're making for less parameters for PAs.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
```diff
@@ -88,7 +95,7 @@ public EntityBullet(Point3D position, Point3D motion, RotationMatrix orientation
         this.gun = gun;
         this.bulletNumber = bulletNumber;
         gun.currentBullet = this;
-        this.isBomb = gun.definition.gun.muzzleVelocity == 0;
+        this.isBomb = gun.muzzleVelocityVar.currentValue == 0;
```
DonBruce64: Good catch on this.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
```diff
@@ -187,13 +195,29 @@ public void update() {
...
             //Add motion requested watch tick we are accelerating.
             boolean notAcceleratingYet = definition.bullet.accelerationDelay != 0 && ticksExisted < definition.bullet.accelerationDelay;
             if (velocityToAddEachTick != 0 && !notAcceleratingYet && ticksExisted - definition.bullet.accelerationDelay < definition.bullet.accelerationTime) {
                 motionToAddEachTick.set(0, 0, velocityToAddEachTick).rotate(orientation);
                 motion.add(motionToAddEachTick);
             }
 
+            //Add turbulent forces, if we're unstable and once we've counted down our curve and accel timers.
+            if ((notAcceleratingYet || ticksExisted > (definition.bullet.accelerationDelay + definition.bullet.accelerationTime)) && ticksExisted > definition.bullet.trajectoryCurveDelay && definition.bullet.trajectoryCurveStability <= Math.random()) {
+                this.trajectoryCurvatureInvertX = Math.random() >= 0.5 ? true : false;
```
DonBruce64: For booleans, don't do ? true : false.  That's redundant.  Booleans will self-assign as boolean.  What you're doing is the old if (condition) == true, and that makes programmers cry.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONPart.java
mklave4d6v8: Actually, I had intended for -1 to work as damage "rampup" (where you deal less damage when at closer range), thus why I set it up this way. I *could* LC/Default it to 1 and have a boolean or something of the sort to invert it, if having falloff and rampup in mind would work better this way...

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
mklave4d6v8: I was eepy sleepy when I wrote that, sowwy. ):

mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONBullet.java
mklave4d6v8: We *do technically* reduce damage for both vehicles and entities when bullets slow down, but I wanted a solution for damage falloff that didn't require slowing down boolet, and it also got very wonky when trying to speed up bullets for rampup. 

That being said, can reword this to better reflect how the damage formulae works.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
mklave4d6v8: :3

I also caught us not accounting for config values if boolets are marked as `isHeat`, on line 343 where we handle damage calculations. It seemed like it wasn't right in testing, though I'd never been one to tweak my boolets using config all too often.
https://github.com/DonBruce64/MinecraftTransportSimulator/pull/2116/changes#diff-5e3c98a79b5c7255cf93e6314db8a52dac6af790fef8ef1e7b1c9cd163e4b6bdR343

mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONPart.java
DonBruce64: Ahh, that makes more sense then.  You're using it as a modifier, not a factor.  That's just fine as a 0 is default positive is falloff, negative means gain.  I'm good as-is with this!

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
DonBruce64: I think the heat bullets just do specific damage values?  But that's more a note for @dldev32 to ponder for perhaps a quick PR.  I don't know if that would require changes down the road or if that's defined behavior as I'm not familiar with heat stuff.

mccore/src/main/java/minecrafttransportsimulator/jsondefs/JSONBullet.java
DonBruce64: Re-wording accepted, and understood about wanting to change boolet behavior.  Sometimes, parameter is easier than physics.

PR #2120, opened 2026-06-09
Added "show hover text" config thing instead of help button on panels

SnailPerson:
@DonBruce64 
As suggested, instead of having the toggle for the hover text on the panel, it instead is a client config option


mcinterfaceforge1122/src/main/resources/assets/mts/recipes/charger.json
```diff
@@ -26,4 +26,4 @@
     "result": {
         "item": "mts:mts.charger"
     }
-}
\ No newline at end of file
+}
```
DonBruce64, line 29: I don't think there was a change here?  I think you wanna delete this from the commit.

DonBruce64 (changes requested):

mcinterfaceforge1122/src/main/resources/assets/mts/recipes/charger.json
SnailPerson: @DonBruce64 you can merge, i always pick a random file im which i add a space and delete it again. That way. Github creates a new branch for me to work in for a PR.

DonBruce64: That makes for bad change tracking.  I would request in the future you pick a file that you know you want to work in first.

SnailPerson: fair portion

PR #2124, opened 2026-07-06
Fixed faulty wrapWidth for signs and such

SnailPerson:
wrapWidth currently has inconsistent units:

Normal line wrapping: 1 becomes one block = 16 model pixels.
autoScale: 1 remains one pixel.
Sign-editor preview: treats it as pixels.

The old code does this:
scale /= 16 converts text scale from pixels to world/block coordinates.
wrapWidth /= scale compensates wrapping for text scale.
The problem is that wrapWidth itself was never converted to world coordinates. Dividing a pixel width by a block-space scale introduces the extra factor of 16.

So one could never do sub pixel non scaling line breaks on signs for example

World line wrapping now treats wrapWidth as pixels, matching auto-scaling and the sign-editor preview. Centered wrapped text was corrected as part of the same conversion.


DonBruce64:
@SnailPerson Interesting.  Does this also fix the funny wrapping of the hand scanner?  I thought wrapWidth was working properly for all signs, though I did see funny things when there were small pixel deltas.

SnailPerson:
> @SnailPerson Interesting. Does this also fix the funny wrapping of the hand scanner? I thought wrapWidth was working properly for all signs, though I did see funny things when there were small pixel deltas.

ill check, give me.a sec

SnailPerson:
I guess it kind of works? Was it different before?
Instruments def look better I think

Manuals seem unaffected

image

image

DonBruce64:
Eh, it's no worse off than before.  Still some issues with some maths somewhere that makes the thing not go long enough on the background, but it works well enough and fixes what ails the signs.

PR #2127, opened 2026-07-12
Made simpleThrottle capable of shifting manual transmissions

mklave4d6v8:
Copilot title: "Enhance gear shifting logic and update shiftCooldown visibility"

simpleThrottle will now try to shift manual engines when they are at 87.5% of their revLimitRPM*
Engines will now always countdown the shiftCooldown, even if they are manual.

Probably messy but I'm a very sad clanker right now. 

5kAUC0ng

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleE_Powered.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/AEntityVehicleE_Powered.java
@@ -47,6 +47,7 @@ public abstract class AEntityVehicleE_Powered extends AEntityVehicleD_Moving {
 	public final ComputedVariable reverseThrustVar;
 	public final ComputedVariable electricUsageVar;
 	public final ComputedVariable batteryCapacityVar;
+    public final ComputedVariable isSimpleThrottleVar;
     public static final double MAX_THROTTLE = 1.0D;
 
     //Internal states.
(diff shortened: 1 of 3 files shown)


SnailPerson:
@DonBruce64 this change has been discussed among PAs, it appears that simple throttle does not shift manual engines, which makes players often floor the car in first gear, despite technically opting to not use it by having simple throttle enabled.
Therefore, the development of such change has been discussed and approved. IV shouldn't gatekeep its content, those that want it use it, those that dont, dont. Everyone is happy.

DonBruce64:
@SnailPerson totally agree and understand this. ST is designed to make things accessible for average user.  Agree with the thoughts on this change.

@boot2big as to implementation, this works on clients but won't work on servers.  Control System is only on clients and will see it's time to shift up, but will send a packet every tick to the server for a shift up request.  This will lead to excess packets arriving on the server and potentially multiple shift operations.  It works on SP due to packets arriving and coming back instantly.

To make this work properly you'll need to have the system know if the controller of a vehicle has simple throttle.  On the vehicle, make a new variable called simpleThrotleVar.  Then on the client in the control vehicle generic section check that variable against the user's config.  If not matching, send request to set it to such.  Then make the engine reference this simpleThrottleVar as if it was an isAutomatic, doing an || in places for shifting.   Can make variables back to private in the engine in this case too.  This ensures that the variable gets set by control system on the vehicle for all vehicles, but you're only setting the variable and leaving engines to shift as if they were automatics, which is what they would do with ST anyways.  Less code in general that way too, as shifting is only in one spot and not a little side one in the controls system.

DonBruce64 (changes requested):
See long form comment in chat.

mklave4d6v8:
D'oh! Gotcha, I'll get on with unifying simpleThrottle with both the server and vehicle, and having a net request to sync said config with the server. Thank you Don!! ^-^

mklave4d6v8:
image
Oh, sure. Yaaay. I thought I told GitHub to put all my changes into my `easyshift` branch instead of my forked `main` branch... thank you Git, very cool.

DonBruce64 (approved):

PR #2128, opened 2026-07-12
minor client config description tweaks

SnailPerson:
simple throttle will get new functionality with boot2bigs commit. 
Also the note that it activates the parking brake when getting of is fake news and false information, so I rectified that. Also shortened the description. googol gaga

Also specified for useShifter that its about a physical real world shifter, because people cant read 

---ezgif com-resize


SnailPerson:
@DonBruce64 take a look lease

Also idk why the diffs mark half the json again, I didn't do nothing there kew

DonBruce64 (approved):

PR #2134, opened 2026-07-21
F1 now hides IVs GUI elements properly

SnailPerson:
Pressing F1 now hides IV’s vehicle HUD, screen instruments, reticles, and overlay text. 
Interactive inventory-style screens remain visible.


SnailPerson:
@DonBruce64 googoo gaga

PR #2142, opened 2026-08-06
Fixed a bug with handheld guns desync

dldev32:
Judging by other people's tests, this fix overcomes this bug with 99% probability, although I personally encountered it very rarely.


DonBruce64:
@dldev32 So just so I understand this, is the gun de-syncing from the player position, or is the gun just getting confused?  Ideally we'd check the player list first, and if the player can't be found, then look up by UUID as a fallback.  Player list is smaller and faster.  So if it's just the matter of it not finding the player I think that would be preferable and more performant.  The entity object returend from either list should be identical.

dldev32:
> @dldev32 So just so I understand this, is the gun de-syncing from the player position, or is the gun just getting confused? Ideally we'd check the player list first, and if the player can't be found, then look up by UUID as a fallback. Player list is smaller and faster. So if it's just the matter of it not finding the player I think that would be preferable and more performant. The entity object returend from either list should be identical.

Yes, you're right. I fixed the issue as reported.

DonBruce64:
Minor techincality, but request 
`foundPlayer = foundEntity instanceof IWrapperPlayer ? (IWrapperPlayer) foundEntity : null;`

be changed to 
```
if(foundEntity instanceof IWrapperPlayer){
   foundPlayer =  (IWrapperPlayer) foundEntity;
}
```

One shouldn't be assigning null to something that's already known to be null going into that if block.  It's an extra operation that's not needed.  That little thing and he'll be good to go.

dldev32:
> Minor techincality, but request `foundPlayer = foundEntity instanceof IWrapperPlayer ? (IWrapperPlayer) foundEntity : null;`
> 
> be changed to
> 
> ```
> if(foundEntity instanceof IWrapperPlayer){
>    foundPlayer =  (IWrapperPlayer) foundEntity;
> }
> ```
> 
> One shouldn't be assigning null to something that's already known to be null going into that if block. It's an extra operation that's not needed. That little thing and he'll be good to go.

Done.

PR #2143, opened 2026-08-06
Added support for OBB collisions

dldev32:
Key features:
1) Added the `length` parameter for collisions, finally allowing for fully three-axis collision dimensions;
2) Added collision rotation based on animations;
3) Added the `rot` parameter for setting the initial rotation of collisions;
4) Added calculation of the effective armor thickness. Its value is taken from the `armorPenetration` parameter + armor slope angles. For correct operation, you only need to specify the actual armor thickness in millimeters;
5) Added the `volumetricArmor` parameter, which allows using the collision box dimensions instead of `armorThickness`;
6) Added the `explosionResistance` parameter, which specifies how much a given box reduces the impact of an explosion, protecting all entities and parts from the explosion behind it;
7) Optionally, you can create collisions directly from the model mesh, the only limitation is the mesh shape, allowing you to use only cubes.

_*Note: The principle of this type of collision works only internally in the mod and is not a collider for entities from МС or other mods.*_

> Please note that OBB collisions only work with certain collision types: `BULLET`, `ATTACK`, `CLICK`, and `VEHICLE`.

An updated plugin for working with new collisions will be ready later.


SnailPerson:
dont merge yet, I want this tested properly

DonBruce64:
le noted.

LauraDarkez:
having tested this pretty well, i want to say its working really well, the only problem is that it doesnt collide with players or blocks, but dl has made it apparent that it would be better to get this in as is, so that collisions with players, blocks and entities can be properly overhauled for all collision variants later

LauraDarkez:
note: i have not tested point 7, 6, or 5, but i dont see any issues

dldev32:
@DonBruce64 are you going to push this PR? I’ve done everything necessary in this branch.

DonBruce64:
@dldev32 I will once @SnailPerson tests since he said he wanted to test.  Please hold till he tests.

PR #2147, opened 2026-08-14
Fixed bugs when flying in first-person camera and improved helicopter controls

dldev32:


DonBruce64:
@dldev32  before I review this, was it intentional that the coremod and such for MC 12 was delted?

dldev32:
> @dldev32 before I review this, was it intentional that the coremod and such for MC 12 was delted?

Yes, they were needed for the first-person cursor to work when the camera edits were used. I returned the MC camera system, because it cannot be done otherwise.

DonBruce64:
@dldev32 ` returned the MC camera system, because it cannot be done otherwise.` I'm confused about this statement.  I get you saying that you needed it to have the camera edits work, but you don't need it now for some reason?  Just want to understand here.

dldev32:
> @dldev32 ` returned the MC camera system, because it cannot be done otherwise.` I'm confused about this statement. I get you saying that you needed it to have the camera edits work, but you don't need it now for some reason? Just want to understand here.

Before this PR, the arcade mode, when playing from the first person, used the same system as from the third person and this created a lot of issues, so I had to remove it.

DonBruce64:
Ahhh, so you reverted some of the stuff you added in 6c0ebe94c58b2287411a6029b1a1b57cd82a4733 then yes?

dldev32:
> Ahhh, so you reverted some of the stuff you added in [6c0ebe9](https://github.com/DonBruce64/MinecraftTransportSimulator/commit/6c0ebe94c58b2287411a6029b1a1b57cd82a4733) then yes?

True.

mccore/src/main/java/minecrafttransportsimulator/packets/instances/PacketVehicleRotorControl.java
```diff
@@ -6,9 +6,9 @@
 import minecrafttransportsimulator.packets.components.APacketEntity;
 
 /**
- * Synchronizes the world-space horizontal direction used to vector helicopter
- * rotor thrust in arcade mode.  Both components travel in one packet so they
- * cannot be mixed between input ticks.
+ * Reserved compatibility packet for the former direct rotor-vector arcade control.
```
DonBruce64: only comment here is we shouldn't be keeping old packets around for registration.  When a mod update releases, it's a new version and forces both the client and server to update.  This ensures both get the updated IDs for packets.

DonBruce64 (changes requested):

dldev32:
@DonBruce64 done.

DonBruce64:
Merged.  Again though, do try to split things into their own PRs.  This was two changes in one.

PR #2150, opened 2026-08-15
Fixed bullet hits on entities

dldev32:
If you noticed earlier, with hitbox visibility enabled, or using particles, you could see that they were spawning at the entity's feet, which is why the bullet's position and orientation are now used instead of the entity's position.

### DIFF
--- a/mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
+++ b/mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
@@ -63,7 +63,6 @@ public class EntityBullet extends AEntityD_Definable<JSONBullet> {
     public boolean waitingOnActionPacket;
     private int impactDespawnTimer = -1;
     private Point3D targetPosition;
-    private final Point3D helperPoint = new Point3D();
     public double targetDistance;
     private double distanceTraveled;
     public double armorPenetrated;
@@ -310,6 +309,11 @@ public class EntityBullet extends AEntityD_Definable<JSONBullet> {
                 }
             }
 
+            //Set the angles to match the current motion before collision checks so impact orientation is current.
+            if (!isBomb && (definition.bullet.accelerationDelay == 0 || ticksExisted > definition.bullet.accelerationDelay)) {
+                orientation.setToVector(motion, true);
+            }
+
             //Long-range bullets and those fired by NPCs do checks on server only, all others do so on clients.
             //We only do client checks on the primary client, not all clients.
             if (((definition.bullet.isLongRange || !(gun.lastController instanceof IWrapperPlayer)) ^ world.isClient()) && (!world.isClient() || InterfaceManager.clientInterface.getClientPlayer().getID().equals(gun.lastController.getID()))) {
@@ -322,7 +326,9 @@ public class EntityBullet extends AEntityD_Definable<JSONBullet> {
                 AEntityF_Multipart<?> hitMultipart = null;
                 Collection<BoundingBoxHitResult> hitMultipartBoxes = null;
                 IWrapperEntity hitExternalEntity = null;
+                Point3D hitExternalEntityPosition = null;
                 hitBlock = world.getBlockHit(position, motion);
+                Point3D endPoint = position.copy().add(motion);
                 
                 //Check for collided external entities.
                 List<IWrapperEntity> attackedEntities = world.attackEntities(damage, motion, true);
@@ -330,25 +336,26 @@ public class EntityBullet extends AEntityD_Definable<JSONBullet> {
                     //Check to make sure we don't hit our controller.
                     //This can happen with hand-held guns at speed.
                     if (!entity.equals(gun.lastController)) {
+                        BoundingBox entityBounds = entity.getBounds();
+                        BoundingBoxHitResult entityHit = entityBounds.getIntersection(position, endPoint);
+                        Point3D entityHitPosition = entityHit != null ? entityHit.position : entityBounds.globalCenter.copy();
+
                         //Make sure there's not a block in the way.
-                        if (hitBlock != null && position.isFirstCloserThanSecond(hitBlock.hitPosition, entity.getPosition())) {
+                        if (hitBlock != null && position.isFirstCloserThanSecond(hitBlock.hitPosition, entityHitPosition)) {
                             continue;
                         }
 
                         //Check if already-found entity is closer.
-                        if (hitExternalEntity != null) {
-                            //Need to use helper here since the position object will be re-used on next call to other entity.
-                            helperPoint.set(hitExternalEntity.getPosition());
-                            if (position.isFirstCloserThanSecond(helperPoint, entity.getPosition())) {
-                                continue;
-                            }
+                        if (hitExternalEntityPosition != null && position.isFirstCloserThanSecond(hitExternalEntityPosition, entityHitPosition)) {
+                            continue;
                         }
                         hitExternalEntity = entity;
+                        hitExternalEntityPosition = entityHitPosition;
                     }
                 }
 
-                //If we hit a entity, and we have a block hit, we need to discard the block.
-                //The only way tne entity could be hit is if it was in front of the block, and thus the block shouldn't be hit.
+                //If we hit an entity, and we have a block hit, we need to discard the block.
+                //The only way the entity could be hit is if it was in front of the block, and thus the block shouldn't be hit.
                 if (hitExternalEntity != null) {
                     hitBlock = null;
                 }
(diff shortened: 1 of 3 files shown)


mccore/src/main/java/minecrafttransportsimulator/packets/instances/PacketEntityBulletHitGeneric.java
```diff
@@ -39,6 +46,7 @@ public PacketEntityBulletHitGeneric(ByteBuf buf) {
         this.position = readPoint3dFromBuffer(buf);
         this.hitType = HitType.values()[buf.readByte()];
         this.hitSide = Axis.values()[buf.readByte()];
+        this.orientation = buf.readableBytes() >= 9 * Double.BYTES ? readRotationMatrixFromBuffer(buf) : null;
```
DonBruce64: Agree with the philospphy of this change.  Only ask is how you send the packet data.  Right now you're sending 9 Doubles for the whole matrix.  For your case, you don't need the whole transform in its entirety.  You can convert it to angles and save those as a Point3d type for 3 Doubles, then re-create the matrix with those doubles on the other end.  It might not be the same state, but the orientation of the bullet at hit-time is what you're looking for, not the ability to rotate it further or suss out how it got to those angles.

This saves 6 Double values per packet, at the expense of some CPU cycles which really aren't much of a bottleneck in this day and age.  I'd be much more concerned about network traffic instead.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/packets/instances/PacketEntityBulletHitGeneric.java
dldev32: > Agree with the philospphy of this change. Only ask is how you send the packet data. Right now you're sending 9 Doubles for the whole matrix. For your case, you don't need the whole transform in its entirety. You can convert it to angles and save those as a Point3d type for 3 Doubles, then re-create the matrix with those doubles on the other end. It might not be the same state, but the orientation of the bullet at hit-time is what you're looking for, not the ability to rotate it further or suss out how it got to those angles.
> 
> This saves 6 Double values per packet, at the expense of some CPU cycles which really aren't much of a bottleneck in this day and age. I'd be much more concerned about network traffic instead.

Done.

mccore/src/main/java/minecrafttransportsimulator/packets/instances/PacketEntityBulletHitGeneric.java
```diff
@@ -39,6 +46,7 @@ public PacketEntityBulletHitGeneric(ByteBuf buf) {
         this.position = readPoint3dFromBuffer(buf);
         this.hitType = HitType.values()[buf.readByte()];
         this.hitSide = Axis.values()[buf.readByte()];
+        this.orientationAngles = buf.readableBytes() >= 3 * Double.BYTES ? readPoint3dFromBuffer(buf) : null;
```
DonBruce64: I don't think you need this check here?  Is the orientationAngles ever null?  It shouldn't be since bullets will always have orientation, and thus you shouldn't need to check if you have bytes remaining.

mccore/src/main/java/minecrafttransportsimulator/packets/instances/PacketEntityBulletHitGeneric.java
```diff
@@ -49,10 +57,17 @@ public void writeToBuffer(ByteBuf buf) {
         writePoint3dToBuffer(position, buf);
         buf.writeByte(hitType.ordinal());
         buf.writeByte(hitSide.ordinal());
+        writePoint3dToBuffer(orientationAngles, buf);
     }
 
     @Override
     public void handle(AWrapperWorld world) {
-        EntityBullet.performGenericHitLogic(world.getBulletGun(gunID), bulletNumber, position, hitSide, hitType);
+        RotationMatrix orientation = orientationAngles != null ? new RotationMatrix().setToAngles(orientationAngles) : null;
```
DonBruce64: Same comment about null angles, though see above comment in the bullet code.

mccore/src/main/java/minecrafttransportsimulator/packets/instances/PacketEntityBulletHitGeneric.java
```diff
@@ -20,14 +21,20 @@ public class PacketEntityBulletHitGeneric extends APacketBase {
...
+        this(gun, bulletNumber, position, getBulletOrientation(gun, bulletNumber), hitSide, hitType);
+    }
+
+    public PacketEntityBulletHitGeneric(PartGun gun, int bulletNumber, Point3D position, RotationMatrix orientation, Axis hitSide, HitType hitType) {
         super(null);
         this.gunID = gun.uniqueUUID;
         this.bulletNumber = bulletNumber;
-        this.position = position;
+        this.position = position.copy();
+        this.orientationAngles = new RotationMatrix().set(orientation).convertToAngles().copy();
```
DonBruce64: So for this you don't want to do this this way.  You're creating a new matrix, setting it to the value of the orientation, converting that new matrix to angle form, then copying those angles.  You should just convert the existing orientation matrix to angle form and use those angles since it'll be the same resulting angles.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
```diff
@@ -639,12 +640,18 @@ public static void performBlockHitLogic(PartGun gun, int bulletNumber, Point3D b
...
     public static void performGenericHitLogic(PartGun gun, int bulletNumber, Point3D position, Axis hitSide, HitType hitType) {
+        EntityBullet bullet = gun.world.getBullet(gun.uniqueUUID, bulletNumber);
+        performGenericHitLogic(gun, bulletNumber, position, bullet != null ? bullet.orientation : null, hitSide, hitType);
+    }
+
+    public static void performGenericHitLogic(PartGun gun, int bulletNumber, Point3D position, RotationMatrix orientation, Axis hitSide, HitType hitType) {
         //Query up return packets first.  This ensures that we get to do this generic logic which spawns particles on clients before
         //any block-breaking packets arrive.
+        EntityBullet bullet = gun.world.getBullet(gun.uniqueUUID, bulletNumber);
+        RotationMatrix impactOrientation = orientation != null ? orientation : bullet != null ? bullet.orientation : new RotationMatrix();
```
DonBruce64: Orientation shouldn't be null, clean up please.

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
```diff
@@ -639,12 +640,18 @@ public static void performBlockHitLogic(PartGun gun, int bulletNumber, Point3D b
     }
 
     public static void performGenericHitLogic(PartGun gun, int bulletNumber, Point3D position, Axis hitSide, HitType hitType) {
+        EntityBullet bullet = gun.world.getBullet(gun.uniqueUUID, bulletNumber);
+        performGenericHitLogic(gun, bulletNumber, position, bullet != null ? bullet.orientation : null, hitSide, hitType);
```
DonBruce64: I see here is where all your null orientation variables come from.  Rather than checking if they are null and doing all the code for that, it would be better to use this check here to just make a new RotationMatrix object to feed all other functions if the bullet can't be found.  That way it'll use the default value cascading down through all functions vs you needing to check for it.

Or, perhaps you've found out what I've just thought of.  That the orientation for a bullet may not always be the same on clients and servers and you want to ensure the unique client values are used?  Or am I over-thinking things here?

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/entities/instances/EntityBullet.java
dldev32: > you want to ensure the unique client values are used?

I guess so. I just wanted to make sure that everything works like clockwork, but most likely you are right, I will try to work on the code.

dldev32:
@DonBruce64 maybe this time?

mccore/src/main/java/minecrafttransportsimulator/packets/instances/PacketEntityBulletHitGeneric.java
```diff
@@ -20,14 +21,20 @@ public class PacketEntityBulletHitGeneric extends APacketBase {
     private final UUID gunID;
     private final int bulletNumber;
     private final Point3D position;
+    private final Point3D orientationAngles;
     private final Axis hitSide;
     private final HitType hitType;
 
     public PacketEntityBulletHitGeneric(PartGun gun, int bulletNumber, Point3D position, Axis hitSide, HitType hitType) {
```
DonBruce64: I don't think this is used anywhere?  If so, why not just feed orientation directly?  Also, if it's not used, you don't need the static function down there.

DonBruce64 (changes requested):

mccore/src/main/java/minecrafttransportsimulator/packets/instances/PacketEntityBulletHitGeneric.java
dldev32: > I don't think this is used anywhere? If so, why not just feed orientation directly? Also, if it's not used, you don't need the static function down there.

Done.

PR #2151, opened 2026-08-15
Minor tweaks to particles

dldev32:
In this PR the following was done:
* `ATTACHED_Z` for `spawningOrientation` - Particle spawns attached to the entity that spawned it, and moves with that entity, but the texture will be rotated for the Z axis;
* `TRAIL` for `spawningOrientation` - Particle spawns as a segment between the previous and current spawn positions, with the horizontal texture axis aligned to the local Z-axis;
* `ZAXIS` for `renderingOrientation` - Particle rotates to face the player, with particle rotation constrained to the Z-axis relative to the player camera;
* The `scale` parameter has been changed to use `Point3D`, allowing for scaling the particle sprite along three axes. `LegacyCompat` still allows the use `float`, maintaining compatibility with older packs.


DonBruce64:
Seems legit to me.  @SnailPerson you're the one on particle crack here.  If all isn't well, lemme know in a week and I'll halt merging.  Otherwise, I'll put er in.

DonBruce64:
@dldev32 do remind me in a week if snail hasn't blessed and I'll merge.

SnailPerson:
@DonBruce64 seems fine, I played with this and it worked.

@dldev32 can you also add the heat haze particle effect for the supporting versions? 
Since it doesn't work on older versions, just have a particle not spawn when it has isheathaze true while being on lower versions if possible

dldev32:
@DonBruce64 you heard what @SnailPerson wrote, so you can merge the PR so that I can continue working in this direction.

PR #2157, opened 2026-08-25
Improved the system for installing parts and placing vehicles

dldev32:
Here's what's been done:
* Holoboxes has been removed;
* Added part slot markers for all part types;
* Added the `installTime` parameter for part installation time.
* Added the `removeTime` parameter for part remove time.
* Added the `deployTime` parameter for vehicle deployment time.
* Added the `packTime` parameter for vehicle packing time.
* Added the `markerName` parameter that defines name of the part slot marker.
* Added the `markerIcon` parameter that defines icon of the part slot marker.


DonBruce64:
@SnailPerson  this okay with you?  It is farly cleaner, I will admit, but probably should be tested.

dldev32:
> @SnailPerson this okay with you? It is farly cleaner, I will admit, but probably should be tested.

I had to use too much code, but it was worth it. I think that's fair.

SnailPerson:
im not sure on the usefulness of the placing feature but the part scanner works fine and is an upgrade over the old system.

Only gripe is that bobbing is wierd with it, it lags behind the actual slots to an extent.
Also @dldev32 before merging, please add the ability to have custom slot names and icons.

SnailPerson:
actually one sec
@dldev32 how does this system handle parts that are stuck behinfd other boxes?
does it take the original slotHeight and width to determine the range of interaction?

dldev32:
> actually one sec @dldev32 how does this system handle parts that are stuck behinfd other boxes? does it take the original slotHeight and width to determine the range of interaction?

Actually, yes. Although I wanted to make it so that the marker itself and its dimensions would be the trigger for placing parts, and also to be able to place parts through collisions of the 'CLICK' type, I was too skeptical that anyone would approve of it.

SnailPerson:
@dldev32 the range of the scanner should be longer, so you see icons from farther away.

LauraDarkez:
> im not sure on the usefulness of the placing feature but the part scanner works fine and is an upgrade over the old system.
> 
> Only gripe is that bobbing is wierd with it, it lags behind the actual slots to an extent. Also @dldev32 before merging, please add the ability to have custom slot names and icons.

i really like the idea of the place time, because it allows for things like installing/uninstalling sounds to be more detailed, along with things like particles, for those who love details. not to mention, installing something heavy like a large bomb should take more time than installing a light bomb

LauraDarkez:
oh, please tell me @dldev32 is the parameter of part install and removal time able to be defined on the part itself as opposed to the slot?

dldev32:
> oh, please tell me @dldev32 is the parameter of part install and removal time able to be defined on the part itself as opposed to the slot?

It's as if everything already exists.

SnailPerson:
doing icons

PR #2160, opened 2026-08-31
Long Range Particle Rendering

SnailPerson:


SnailPerson:
@DonBruce64 what I did is implement long range particles. 
Please check this because the way its implemented works, but its a little hacky.
Basically specific particles are rendered WAY beyond the bounds of normal rendering. Like, 150k blocks if you define it so. 
Useful for plane LODs, flare guns etc


But yeah, make sure everything makes sense, this is experimental

dldev32:
> @DonBruce64 what I did is implement long range particles. Please check this because the way its implemented works, but its a little hacky. Basically specific particles are rendered WAY beyond the bounds of normal rendering. Like, 150k blocks if you define it so. Useful for plane LODs, flare guns etc
> 
> But yeah, make sure everything makes sense, this is experimental

First of all, to ensure this feature works reliably, you need to decouple the particles from the Minecraft entity system. As long as the particles remain entities, you’re going to face major issues—especially if you’re spawning them outside the simulation distance.

PR #2163, opened 2026-09-02
Tweaked Fuel Section Description mtsconfig

SnailPerson:
lowkey made the description a lilt easier to understand and also explaind what tf poenlcy even does

image


SnailPerson:
@DonBruce64 made config easier to read

PR #2172, opened 2026-09-10
Allowed the use of parking brake when simpleThrottle is true, and added the activation of the parking brake when the driver exits the vehicle

dldev32:


DonBruce64:
@SnailPerson you had a PR to remove this from the manual, yes?  do we want this back?

dldev32:
> @SnailPerson you had a PR to remove this from the manual, yes? do we want this back?

He removed them, but players didn't realize this and reported that the parking brake wasn't working, so I fixed it. This might also temporarily mitigate the sliding issue, since the car won't be able to ride on its own after the player exits.

DonBruce64:
Ideally, this should be how it is.  The car goes into park when you exit it so it doesn't roll away or slide.

DonBruce64:
I see that we don't do parking brakes on aircraft, or vehicles with running engines.  That intentional?

SnailPerson:
@DonBruce64  tbh idk what this PR is trying to fix.
The parking brake behaviour is embedded in autostart eng already and you can still pull the handbrake with simple throttle.
None of that was removed.

dldev32:
> @DonBruce64 tbh idk what this PR is trying to fix. The parking brake behaviour is embedded in autostart eng already and you can still pull the handbrake with simple throttle. None of that was removed.

The parking brake engages if both functions are enabled, but this is not stated anywhere.

SnailPerson:
> > @DonBruce64 tbh idk what this PR is trying to fix. The parking brake behaviour is embedded in autostart eng already and you can still pull the handbrake with simple throttle. None of that was removed.
> 
> The parking brake engages if both functions are enabled, but this is not stated anywhere.

it starts because autostartEng is the one triggering it

DonBruce64:
Perhaps it shouldn't be?  Auto start engine I wouldn't think would be what messes with the parking brake.  I would think it would be simpleThrottle.  Perhaps it needs to be removed from the auto-start code?  I would think that this would be a GTAV thing.  You always start cars you enter, but the control scheme can be simple  or complex depending on user preference.

Issue #2169, opened 2026-09-08
Cameras crash servers


LauraDarkez:
because they were moved to serverside operation, the previous system used to make cameras restricted to specific seats not only doesnt work, but actively crashes servers due to being reliant on client variables

there is no currently functional alternative to this, which is why pack authors havent been able to do anything other than either make all cameras accessible to all seats, or keep the system that crashes servers

DonBruce64:
@SnailPerson  @dldev32  I smell PR issues...

LauraDarkez:
actually, this happened before prs, this was from back when you were still head dev. i tried to report it in the discord and it got shut down because it was an intentional change on your part, and that "pack authors shouldnt be using client variables on cameras anymore" or so i remember

dldev32:
Crash report or it didn't happen.

LauraDarkez:
[crash-2026-09-08_15.21.16-server.txt](https://github.com/user-attachments/files/32004998/crash-2026-09-08_15.21.16-server.txt) is this proof enough that it happened? 

[2026-09-08-4.log](https://github.com/user-attachments/files/32005238/2026-09-08-4.log) or do you need this too?

mind you, this does not crash on singleplayer worlds. or essential hosted worlds. only DEDICATED SERVERS

PR #2173, opened 2026-09-10
Added a client-side world check when using gun_controller_firstperson

dldev32:
Fix related to #2169


LauraDarkez:
tested this myself in 1.12 on a dedicated server and it worked like a dream

DonBruce64:
Although I discourage folks from using client variables on things that could be different states between clients and servers, I reckon this works.  Especially since it's for hand-held gun cameras which don't need to sync.  Approved!

PR #2175, opened 2026-09-11
Fixed camera objects position and fixed #2105 dont work

dldev32:


LauraDarkez:
from what i can tell this works, but i have only tested 1.12

DonBruce64:
This is common code, so no need to test other versions.  Thanks @LauraDarkez !

PR #2176, opened 2026-09-11
Fixed hitbox rendering on none modelType

dldev32:


SnailPerson:
@dldev32 please provide context, were hotboxes invisible In f3 for any vehicle with none. modelType?

dldev32:
> @dldev32 please provide context, were hotboxes invisible In f3 for any vehicle with none. modelType?

Yes, the checks indicate that, and the code completely disables rendering anyway, regardless of what the modelType parameter uses.

DonBruce64:
@dldev32 So from what I'm seeing this is to fix issues when the model type is NONE, but you do have some things that you want to render on said model like maybe particles or instruments or whatnot and those were being blocked by the model disabling all rendering? If so, I'm good with this.  Just wanna double-check.

LauraDarkez:
> @dldev32 So from what I'm seeing this is to fix issues when the model type is NONE, but you do have some things that you want to render on said model like maybe particles or instruments or whatnot and those were being blocked by the model disabling all rendering? If so, I'm good with this. Just wanna double-check.

this was mostly meant to allow for the box that shows the part's size/position to appear like any other part, though particles would also be useful i can foresee. if particles werent able to be spawned by none parts then this will absolutely be beneficial as well

DonBruce64:
Ahh, that makes sense.  Le merged!

PR #2178, opened 2026-09-13
Add anim variable allowing to obtain the true bearing to the beacon.

conman180:
expansion on "beacon_direction" variable
older variable is relative bearing
this variable is true bearing to the beacon. 

example:    i am pointed west (270), beacon is also to the west of me
relative bearing is 0 because i am pointed at beacon
true bearing is 270 because beacon is due west of me regardless of which way i am facing


conman180:
irl calculating this is as simple as relative bearing + heading
but code requires clamps to keep it within 0 and 360

conman180:
also @DonBruce64 long time no see since ive been here hope all is well.

DonBruce64:
Been good!  Just been on vacation for a week which has been nice.

PR #2180, opened 2026-09-15
Dynamic Crosshair Config

SnailPerson:
- Added a config toggle to the p menu for toggling the dynamic crosshair. Default is true.
- Redone potion effect indexing, instead of using preset "instant_harming" tags, the system now simply uses actual potion IDs, meaning modded effects are also possible


SnailPerson:
don t merge yet, gotta revert some stuff

SnailPerson:
@DonBruce64 MWERGE MERGE MERGE DING DING DING GAY SEX GAY SEX

DonBruce64:
@SnailPerson I don't see changes for this?  Was that supposed to get up into the files and fix butter?  If so, seperate PR if you could.

Redone potion effect indexing, instead of using preset "instant_harming" tags, the system now simply uses actual potion IDs, meaning modded effects are also possible

